Issue #397, opened 2026-02-21
Optimize particle rendering: GPU compute for 250K+ at 60 FPS


JamesPagetButler:
## Context

Phase 4e stress testing (#302) revealed that the CPU-bound particle rendering approach hits 5 FPS at 252K particles. Industry benchmarks (2025-2026) show GPU-accelerated systems handling 1M+ particles at 60 FPS.

**Current bottleneck:** CPU-side `DrawInstanceGroup.UpdateData()` in `draw_instance.go:315-358` — per-instance transform, AABB culling, and memcpy at 0.7μs/particle.

## Performance Data

| Particles | Current FPS | Target FPS |
|-----------|------------|------------|
| 1,000 | 800+ | 900+ |
| 10,000 | ~200 | 800+ |
| 100,000 | ~15 | 60+ |
| 250,000 | 5.4 | 60+ |
| 1,000,000 | ~1 (est.) | 50+ |

## Dev Team Analysis (Carmack, Muratori, Pike)

**Key finding:** Kaiju already has instanced rendering (`vkCmdDrawIndexed` with instanceCount, `vk_drawing.go:187`) and compute shader infrastructure (`shader_layout.go:202`, `render_id.vk.go:57`). The GPU plumbing exists but isn't wired up for particles.

### Ranked Optimization Strategies

| Rank | Strategy | Effort | Speedup | Combined Target |
|------|----------|--------|---------|----------------|
| 1 | **Compute shader particle update** — move transforms to GPU | 4-6 hrs | 30-50× | 176ms → ~5ms |
| 2 | **Bulk frustum culling** — skip per-instance AABB | 1-2 hrs | 2-3× | Eliminate ~50ms |
| 3 | **Point sprites** — 1 vertex per particle instead of sphere | 2-3 hrs | 3-5× | GPU efficiency |
| 4 | Compress transfer data (position+color only) | 2-3 hrs | 1.5-2× | Reduce memcpy |
| 5 | Indirect rendering (vkCmdDrawIndirect) | 3-4 hrs | 1.5-2× | GPU-driven draws |

**Combined estimate:** Strategies 1+2+3 → 176ms → 10-15ms → **67-100 FPS at 250K**

## Reference Benchmarks (2026)

- GPU-accelerated: 100-300× speedup over CPU
- 1M+ particles at 50-60 FPS with basic physics
- 168,600 particles/ms throughput achievable
[...]

Issue #302, opened 2026-02-13
Experiment 03: Phase 4e — Verified Simulation Engine


JamesPagetButler:
## Summary

Provide interactive 3D simulation for human Verification & Validation (V&V) of Experiment 03 (Double-Slit), using the proven Lean oracle as the physics source of truth.

**Parent Issue:** #56 (Experiment 03: Double-Slit - Phase 4: Formal Verification)
**Sprint:** 3

## Requirements

- [ ] Implement experiment-specific Go scene for Double-Slit
- [ ] Load oracle predictions from Phase 4d (#301)
- [ ] V&V workflow: setup verification → behavior verification → results validation
- [ ] Document V&V checklist for Experiment 03

## Acceptance Criteria

- [ ] **AC:** Go simulation scene implemented for Experiment 03
- [ ] **AC:** Physics oracle provides Lean-proven predictions
- [ ] **AC:** Human can manipulate experiment setup parameters
- [ ] **AC:** V&V checklist documented and completed

## References

- Phase 4d: #301 (Verified Differential Testing)
- Phase 4e framework: #245 (closed — infrastructure)
- V&V methodology: `docs/workflows/verification_validation.md`
- Simulation code: `src/simulation/`

JamesPagetButler:
## Phase 4e Implementation Plan — Verified Simulation Engine for Double-Slit

### Three-Layer Architecture (refined through 3 rounds of Gemini review)

```
┌─────────────────────────┐
│  Go Physics Engine      │  Implements Fraunhofer I(x) formula
│  (drives real-time sim) │  Handles dynamic slider parameters
├─────────────────────────┤
│  Oracle Truth Snapshots  │  ~5 fixed parameter sets from Lean
│  (from JSON)            │  Used ONLY for V&V validation
├─────────────────────────┤
│  V&V Panel              │  Compares Go engine output vs Oracle
│                         │  Active on presets, "Unvalidated" on custom
└─────────────────────────┘
```

### Experimental Grounding

| Experiment | Particle | λ | d | L | Key Result |
|-----------|----------|---|---|---|------------|
| Bach et al. 2013 | electrons | ~50pm | 272nm | 24cm | Single-electron buildup |
| Zeilinger et al. 1988 | neutrons | 1.845nm | 126.5μm | 5.0m | Neutron fringes |
| Tonomura et al. 1989 | electrons | 0.0055nm | biprism | — | Iconic buildup video |
[...]

Issue #393, opened 2026-02-20
Facility exterior: floating island, space elevator, asteroid spaceport


JamesPagetButler:
## Summary

Build the exterior world for the QBP Research Facility. All reference material is documented in `docs/design/facility_location.md`.

### Floating Island (sea level)
- Island geometry: hull/pontoons, vegetation, solar panels
- Ship port with small dock (Brass/Bronze bollards)
- Helipad with Ochre landing circle
- Facility exterior: brass-framed greenhouse-style structure

### Space Elevator
- Brass/bronze lattice ground tower
- Carbon nanotube tether (visible in skybox)
- Glass-and-solar climber pod (Victorian lift cage aesthetic)
- LEO and GEO waypoint stations

### Asteroid Spaceport (counterweight, ~100,000 km)
- Captured ~500m carbonaceous chondrite
- 3-4 hangars carved into rock with brass rolling doors
- Central spine (tether anchor / climber arrival)
- Observation dome (glass + brass, sunward face)
- Docked research vessels, solar-sail shuttles
- Amber guide strips, Ochre interior lighting
- Slow rotation for partial artificial gravity

### Undersea Elevator
- Rigid shaft from island hull to trench floor
- Trench floor observatory at 10,935m depth
- Brass diving bell / bathyscaphe aesthetic

## Location

Centered on Challenger Deep: 11°22.4′N 142°35.5′E (Mariana Trench, western Pacific).

## Design Reference

- Color palette: `docs/design-system.md` (Solarpunk)
- Full spatial layouts and cross-sections: `docs/design/facility_location.md`

## Notes

- None of this is needed for Sprint 3 — scientist starts inside the hallway
- Could serve as: arrival sequence, loading screen transitions, skybox elements, future experiment locations (space-based labs on the asteroid)
- Priority: low — build when the interior experiments are solid

---

- [ ] **AC:** Floating island exterior visible from a viewpoint
- [ ] **AC:** Space elevator tower + tether visible in skybox
- [ ] **AC:** Asteroid spaceport explorable (hangars, obs dome)
- [ ] **AC:** Undersea elevator / trench observatory explorable
- [ ] **AC:** Arrival sequence (how the scientist gets to the facility)

JamesPagetButler:
## Added: Undersea Elevator — Research Stations & Sub Base

5 research stations at ocean zone boundaries + trench floor sub base added to `docs/design/facility_location.md`.

| Station | Depth | Zone |
|---------|-------|------|
[...]

Issue #301, opened 2026-02-13
Experiment 03: Phase 4d — Verified Differential Testing


JamesPagetButler:
## Summary

Validate that the Python simulation for Experiment 03 (Double-Slit) matches the proven-correct Lean oracle. Uses the Cedar pattern (Amazon) for mathematical confidence that published results are computed correctly.

**Parent Issue:** #56 (Experiment 03: Double-Slit - Phase 4: Formal Verification)
**Sprint:** 3

## Requirements

- [ ] Create experiment-specific Float oracle in `proofs/QBP/Oracle/`
- [ ] Generate JSON test predictions for Experiment 03
- [ ] Run differential harness against `qphysics.py` (10,000+ cases)
- [ ] CI passes with 0 divergences

## Acceptance Criteria

- [ ] **AC:** Lean Float oracle compiles for Experiment 03
- [ ] **AC:** JSON export produces test vectors covering edge cases
- [ ] **AC:** Differential harness detects intentionally-introduced bugs
- [ ] **AC:** CI pipeline passes with 0 divergences for Experiment 03

## References

- Phase 4a: Formal Proof for this experiment
- Phase 4d framework: #242 (closed — infrastructure)
- Differential test harness: `scripts/differential_test.py`
- CONTRIBUTING.md Phase 4d section

PR #401, opened 2026-02-22
feat: Phase 4e — Kaiju Virtual Lab with scientist UX (#302)

JamesPagetButler:
## Summary

- **Full Kaiju-based virtual physics lab** for Experiment 03 (Double-Slit) — scientist sits at a desk with monitors, controls, and instruments
- **NASA/F-35 style control panel** with grouped buttons (Experiment, Parameters, Presets), color-coded by function
- **CPU-rasterized monitor text** on 3D surfaces (solved backface culling + frustum culling + alpha discard issues)
- **Particle cap system** (default 25K) with cycling, prevents FPS death spiral
- **V&V checklist** with auto-check criteria and manual verification items
- **Histogram monitor** showing real-time detector hit distribution
- **FPS camera** with WASD movement, mouse look, seated/standing positions

### Rendering fixes (latest commits)
- Custom `-Z` facing quad mesh (`newMeshQuadNegZ`) to solve backface culling — standard Kaiju quads face +Z, invisible from camera at Z=-6.5
- Custom `+Y` facing quad mesh (`newMeshQuadPosY`) for flat desk surface labels
- `alwaysVisibleCuller` to bypass frustum culling for zero-depth-AABB flat quads
- Opaque label backgrounds (alpha=1.0) to survive engine's 0.999 alpha discard threshold
- Increased text-to-monitor Z-offset (0.015m → 0.04m) to prevent Z-fighting

### Known issues (documented on #302)
- Text visibility from seated position needs James's visual verification after latest fixes
- GPU particle optimization deferred to #397 (research complete, plan posted)

### Also includes
- Facility design documents (Gemini-led: location, central hub, hallway, exterior)
- Paget persona (Lead Bio-Synthesist)
- Environmental conditions added to ground truth schema
- Oppenheimer strategic review #001

## Test plan

- [ ] Build: `cd src/bakeoff/kaiju/engine/src && go build -tags '!editor' -o ../../kaiju-ds .`
- [ ] Run: `VK_ICD_FILENAMES=/etc/vulkan/icd.d/amd_icd64.json QBP_LAB=1 LD_LIBRARY_PATH=../../local_libs ../../kaiju-ds`
- [ ] Press Q to sit — verify monitor text is readable from seated position
- [ ] Verify desk labels ("EXPERIMENT", "PARAMETERS", "PRESETS") visible on desk
- [ ] Click controls via crosshair — verify Start/Stop, Rate, Presets work
- [ ] Verify particle cap prevents FPS death spiral (stays >20 FPS at 25K)
- [ ] Verify histogram updates in real-time on right monitor

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -85,3 +85,15 @@ This log tracks work done locally by Gemini that has not yet been pushed to the 
 - Formalized the **Gemini-Led PR Creation** workflow, where Gemini is now responsible for creating PRs.
 - Added **Guide Posts for Emergent Phenomena** to `paper/quaternion_physics.md`.
 - Created foundational persona documentation for Feynman and Furey review roles.
+
+---
+
+### **Entry: 2026-02-22 (Session 1)**
+
+**Task:** Synthesize Architectural Vision for the Central Hub ("The Nexus").
+**Files Modified:** `docs/design/facility_central_hub.md`
+**Changes:**
+- Integrated ecological aspects, such as Bio-Pontoon Shallows and wildlife habitats.
+- Included 'The Nexus Portals' for transport and logistical practicalities.
+- Developed the 'Seed to Shell' expansion metaphor (logarithmic conchoid spiral around the Celestial Axis).
+- Generated conceptual visualizations representing the 'Seed Form' and 'Conchoid Ascending Shell'.
--- a/SPRINT_STATUS.md
+++ b/SPRINT_STATUS.md
@@ -34,7 +34,7 @@
   - [x] 4b: Proof Review (#260) — CLOSED 2026-02-20. 3 review rounds. V(η) bridge (5 theorems) contextually reviewed. Fraunhofer extracted to QBP.Optics (PR #389). Model A spatial correlation caveat documented (#387). FAULT-S3-007, FAULT-S3-008 logged.
   - [x] 4c: Interactive Proof Visualization (#261) — CLOSED 2026-02-20. PR #391 merged. 39-node proof graph, Atkinson Hyperlegible fonts, 3 review rounds (Red Team + Gemini + human visual).
   - [x] 4d: Verified Differential Testing (#301) — CLOSED 2026-02-20. PR #392 merged. 86 comparisons, 0 divergences. 17 DoubleSlit test vectors, 3 bug detection mutations.
-  - [ ] 4e: Verified Simulation Engine (#302) — IN PROGRESS. Raylib + Kaiju bake-off prototypes built. AMDVLK driver installed for RX 9070 XT.
+  - [ ] 4e: Verified Simulation Engine (#302) — IN PROGRESS. Kaiju prototype: desk controls, monitors, histogram, particle cap (25K), NASA control panel. Text rendering fixes pending James testing (opaque backgrounds, horizontal labels, Z-offset). See #302 comment 2026-02-22.
 - [ ] Phase 5: Publication (#65)
 - [ ] Theory Refinement (#81)
 - [ ] Research Gate: `python scripts/research_gate.py --scope sprint-4 experiment-04`
--- a/docs/Possible Future Experiments/001_far_field_double_slit.md
+++ b/docs/Possible Future Experiments/001_far_field_double_slit.md
@@ -1,0 +1,62 @@
+# 001: Far-Field Double-Slit with Single-Atom Wave Packets
+
+## Motivation
+
+Sprint 3 (Experiment 03, double-slit) produced near-field BPM results only. The simulation propagates ~32 nm — far too short for Fraunhofer far-field conditions (which require mm-scale propagation). As a result:
+
+- **BPM baseline visibility V ≈ 0.553** (vs analytical far-field V = 1.0)
+- The QBP quaternionic coupling signal (V reduction of ~8%) is real but sits on top of a near-field baseline, not a clean far-field fringe pattern
+- We have no far-field QBP prediction to compare against actual double-slit experiments
+
+A far-field simulation would let us predict what a real interferometer would measure, making the QBP theory directly testable.
+
+## Physical Setup — Ketterle Group (MIT, 2025)
+
+A recent paper provides an interesting experimental configuration:
+
+> **"Coherent and Incoherent Light Scattering by Single-Atom Wave Packets"**
+> Fedoseev, Lin, Lu, Lee, Lyu & Ketterle
+> *Physical Review Letters* **135**, 043601 (July 2025)
+> DOI: [10.1103/zwhd-1k2t](https://doi.org/10.1103/zwhd-1k2t)
+
+Key features:
+- Single photons scattered off Heisenberg uncertainty-limited atomic wave packets
+- Ultracold atoms in free space (Mott insulator source)
+- Coherent vs incoherent scattering fraction depends on wave packet size relative to photon wavelength
+- Directly probes atom-photon entanglement and which-way information
+
+This setup is relevant because the QBP predicts that quaternionic coupling modifies the coherent/incoherent partition — precisely what this experiment measures.
+
+## What We'd Simulate
+
+- **Far-field propagation:** Either extend BPM to mm scale (computationally expensive) or use Fourier-optics far-field transform from near-field BPM output
+- **Single-atom wave packets:** Match the Ketterle group's parameters (atom species, wavelength, wave packet size)
+- **Observable:** Coherent scattering fraction as a function of quaternionic coupling strength U₁
+- **Comparison:** QBP prediction vs standard QM prediction vs experimental data (if available)
+
+## Expected QBP Signature
+
+The quaternionic coupling transfers amplitude from the complex component to the quaternionic component via SO(4) rotation. In a far-field interference experiment, this should:
+
+1. Reduce fringe visibility (as seen in near-field Sprint 3 results)
+2. Modify the coherent/incoherent scattering ratio — the quaternionic component acts as an internal degree of freedom that carries which-way information
+3. Produce a coupling-strength-dependent shift in the coherence fraction that differs from standard QM decoherence
+
+## Key References
+
+- Fedoseev et al., PRL 135, 043601 (2025) — DOI: 10.1103/zwhd-1k2t
+- Sprint 3 results: `analysis/03_double_slit/RESULTS.md`
+- Ground truth: `research/03_double_slit_expected_results.md`
+
+## Feasibility Notes
+
+- **BPM capability:** Current BPM handles near-field only. Far-field would require either (a) much larger grid + longer propagation (very expensive), or (b) Fourier transform of near-field output to far-field (moderate effort, new code needed)
+- **Computational cost:** Far-field BPM at mm scale with nm resolution is infeasible (~10⁷ grid points per axis). Fourier-optics approach is the practical path.
+- **Blockers:**
+  - Need to implement Fourier far-field transform
+  - Need Ketterle group's experimental parameters (atom species, trap frequencies, photon wavelength)
+  - Full paper access required (currently paywalled)
+
+## Status
+
+[x] Idea | [ ] Scoped | [ ] Ready to schedule
--- a/docs/Possible Future Experiments/README.md
+++ b/docs/Possible Future Experiments/README.md
@@ -1,0 +1,42 @@
+# Possible Future Experiments
+
+## Purpose
+
+This directory collects experiment ideas that go beyond the current 10-sprint roadmap (experiments 01–09). Each proposal follows the template below so that when we're ready to prioritise, we have enough context to make a decision.
+
+## Proposal Template
+
+When adding a new experiment idea, create a file `NNN_short_name.md` using this structure:
+
+```
+# Title
+
+## Motivation
+Why is this interesting for QBP? What question does it answer?
+
+## Physical Setup
+Brief description of the experimental configuration.
+
+## What We'd Simulate
+BPM parameters, geometry, observables.
+
+## Expected QBP Signature
+What would the quaternionic coupling change relative to standard QM?
+
+## Key References
+Papers, DOIs, relevant prior art.
+
+## Feasibility Notes
+- BPM capability: can our current code handle this?
+- Computational cost: grid size, propagation distance, runtime estimate
+- Blockers: what would need to change first?
+
+## Status
+[ ] Idea | [ ] Scoped | [ ] Ready to schedule
+```
+
+## Current Proposals
+
+| # | Title | Status | File |
+|---|-------|--------|------|
+| 001 | Far-field double-slit with single-atom wave packets | Idea | [001_far_field_double_slit.md](001_far_field_double_slit.md) |
--- a/docs/design/facility_central_hub.md
+++ b/docs/design/facility_central_hub.md
@@ -1,0 +1,54 @@
+# QBP Facility — The Central Hub ("The Nexus")
+
+**Lead Bio-Synthesist:** Paget  
+**Status:** Concept Synthesized (Sprint 3)
+
+## Architectural Synthesis: The Seed and the Shell
+**Lead Bio-Synthesist's Refinement:** Paget  
+**Status:** Synthesis Locked (Sprint 3)
+
+The QBP facility is not merely a laboratory; it is a **cultivated organism** designed to bridge the gap between the rigorous precision of quantum physics and the restorative complexity of biological life. Our design philosophy, the **"Seed to Shell"** model, ensures that the facility's growth is both deeply efficient and profoundly inspiring.
+
+### The Three Pillars of Our Arcology:
+1.  **Ecological Inspiration (The Life-Blood):** We reject the sterile, rigid laboratory. By integrating **Bio-Pontoon Shallows**, **Sub-Sea Architecture**, and **Wildlife Habitats**, we create a porous, living environment. The presence of thriving marine life and the scent of an old-growth forest (our ambient baseline) are not luxuries; they are essential cognitive fuel, providing the psychological grounding required for deep theoretical synthesis.
+2.  **Efficient Practicalities (The Nervous System):** Every organic element serves a logistical purpose. The **Bloom Docks** (petal-like collars) provide weather-protected transitions for heavy cargo, while **Sub-Surface Freight Channels** ensure that the core living spaces remain tranquil. The facility is a masterclass in hidden efficiency—logistics are the "submerged" part of the iceberg, keeping the surface focused on pure research.
+3.  **The Growth Algorithm (The Logarithmic Spiral):** We begin with the **Seed Form (Phase 1)**—a compact, pearlescent disc anchored to the **Celestial Axis** (the space elevator tether). As our research expands, the facility does not just "add rooms"; it undergoes a **Conchoid Spiraling Ascent**. This logarithmic growth weaves hub support and research hallways into a continuous, protective shell that climbs the tether, providing natural structural reinforcement and infinite scalability.
+
+This is our arcology: an iridescent nautilus emerging from the abyss, anchored by the light of the space elevator, and dedicated to the discovery of the universe's fundamental symmetries.
+
+### 1. The Core Spaces (Mind and Body)
+The Nexus is a sprawling, multifaceted structure of vaulted glass domes and sweeping brass arches, housing essential life-support and intellectual spaces:
+- **The Grand Library:** A multi-level, open-air archive bridging physical texts and digital interfaces. It acts as the intellectual core, offering silent spaces for deep theoretical synthesis.
+- **Reflection Gardens:** Indoor and outdoor botanical sanctuaries designed to mimic the tranquility of ancient forests. These areas encourage cognitive wandering and stress recovery.
+- **Living Quarters:** Ergonomic, private sanctuaries built into the bio-pontoon structure, offering restorative views of the deep ocean or the internal gardens.
+- **Recreation & Wellness:** Spaces for physical training, communal dining, and social gathering, ensuring the team remains healthy and connected during long isolation periods above the Challenger Deep.
+
+### 2. Ecological Integration & Biomimicry (The Fluid Organism)
+Moving away from rigid, blocky structures, The Nexus embraces fluid, organic architecture that blurs the line between human engineering and natural habitat:
+- **Bio-Pontoon Shallows & Beaches:** While floating over the abyss, the sprawling bio-pontoons themselves create artificial "shallows." The structure gently slopes into the water, forming pristine, white-sand beaches, secluded swimming coves, and meditative lookout points right at the ocean's surface.
+- **Sub-Sea Architecture:** The living quarters and certain reflection spaces extend seamlessly *below* the waterline, featuring expansive reinforced glass walls that look directly into the vibrant pelagic zone cultivated around the island's roots.
+- **Wildlife Habitats:** The architecture is intentionally porous to native Pacific wildlife. Micro-habitats are engineered into the eaves and pontoons—safe nesting grounds for seabirds, and artificial reef structures to attract and sustain rich marine biodiversity. The presence of thriving wildlife acts as a continuous source of inspiration and psychological grounding for the research team.
+
+### 3. Fluid Transportation & Logistics (The Nexus Portals)
+To ensure the facility remains a functional and connected arcology, The Nexus integrates sophisticated transportation and logistics hubs that blend seamlessly into its organic form:
+- **The Bloom Docks (Arrival & Departure):** Large, petal-like docking collars located at the hub's perimeter. These structures expand and contract to receive transport vessels and heavy equipment, providing stabilized, weather-protected transitions for personnel and cargo.
+- **Recreational Marine Fleet:** Nestled within the calm "Bio-Pontoon Shallows," the facility features integrated slips for sailing yachts, electric hydrofoils, and deep-sea observation submersibles. These areas allow researchers to engage with the ocean for both study and restoration.
+- **Aerial & Drone Platforms:** Discreet, reinforced landing zones atop the vaulted domes for rapid personnel transfer (VTOL) and automated drone delivery. These platforms are camouflaged by the "Fractal Canopy" to maintain the facility's aesthetic profile.
+- **Sub-Surface Freight Channels:** A hidden network of automated conduits beneath the main floorplates. This system moves heavy equipment and supplies from the docking bays directly to the modular experiment branches, ensuring the core living and research spaces remain tranquil and pedestrian-focused.
+
+### 4. The Expansion Metaphor (The Emergent Conchoid Shell)
+The facility is designed not as a static building, but as a **cultivated organism** that grows to meet the demands of the research, climbing the central tether like a massive, spiraling nautilus emerging from the abyss.
+- **The Seed Form (Phase 1):** In its initial stage, the facility rests close to the ocean's surface as a compact, pearlescent disc anchored to the base of the Celestial Axis. This "Seed Form" houses only the core life support, the primary Grand Library, and the initial Sprint 1 & 2 laboratories. It resembles a tightly coiled, nascent shell preparing to open.
+- **Conchoid Spiraling Ascent:** As research demands expand, the island begins its upward growth in a continuous, logarithmic spiral around the Celestial Axis, mimicking the protective and efficient geometry of a deep-sea shell. This "emergent" form provides natural structural reinforcement against oceanic and atmospheric pressures.
+- **Pearlescent Protection:** The outer "shell" of the spiraling tiers is composed of a durable, iridescent bio-composite that protects the internal ecosystems from the harsh maritime environment while allowing filtered, ethereal light to permeate the research and living spaces.
+- **Integrated Hub Support & Research:** This spiraling expansion does not just add raw laboratory space; it seamlessly weaves vital hub support—such as new reflection gardens, localized living quarters, and secondary data nodes—directly alongside the new modular research hallways. Every new tier functions as a semi-autonomous micro-arcology.
+- **Fractal Modularity:** The experiment hallways (each a 10-lab module) branch outward from the spiraling central spine like the limbs of a great tree or a fractal geometric pattern. The higher the facility climbs, the more complex and specialized the branching becomes, yet it always remains connected to the core.
+- **The Central Anchor:** All branches, tiers, and spiraling pathways radiate from and are structurally grounded by the Celestial Axis (the Space Elevator tether). This ensures absolute stability against both oceanic forces and the immense, shifting weight of the expanding arcology.
+
+### 5. Environmental Continuity
+The Nexus maintains the facility-wide **Ambient Baseline**:
+- **Atmosphere:** Old Growth Forest (21.5% $O_2$, Trace Phytoncides)
+- **Temperature:** 22°C (295.15 K)
+- **Pressure:** 101.325 kPa (Stable Sea-Level)
+
+The transition from The Nexus to any experiment wing is mediated by the "Symmetry of Utility" — researchers leave the restorative environment of the Hub, pass through the Gear Room of a specific module, and enter the rigorous, environmentally-controlled space of the experiment hallway.
--- a/docs/design/facility_exterior/paget_concept.md
+++ b/docs/design/facility_exterior/paget_concept.md
@@ -1,0 +1,64 @@
+# QBP Facility Exterior — Paget's Bio-Synthetic Concept
+
+**Lead Bio-Synthesist:** Paget  
+**Team:** Bio-Symmetry Design Group  
+**Issue:** #393 (Facility Exterior)
+
+---
+
+## 1. The Living Anchor (Floating Island)
+
+The island is not "placed" on the sea; it is **cultivated**. The hull is a composite of bio-pontoons and carbon-fiber lattice, grown in the warm Pacific currents of 11°22.4′N.
+
+### Key Features
+- **The Hull:** A non-associative geometric structure that absorbs wave energy to power the base station. Its form is determined by the specific frequency of the Challenger Deep swells.
+- **The Greenhouse Exterior:** Brass-framed, glass-enclosed labs. The "The Living Axiom" dictates that the glass is photo-reactive, changing opacity based on the solar flux.
+- **Vegetation:** Sage and Patina-toned "Solarpunk" greenery that provides both oxygen and psychological grounding for the research team.
+
+---
+
+## 2. The Celestial Axis (Space Elevator)
+
+The space elevator is the **Symmetry of Utility**'s highest expression. Every meter of the 100,000 km tether is a data-conduit and a tension-balancer.
+
+### Key Features
+- **Ground Tower:** A bronze/brass lattice "anchor" that feels like a Victorian clockwork mechanism grown to a massive scale.
+- **The Tether:** A carbon-nanotube ribbon, nearly invisible but for the way it "refracts" the high-altitude light.
+- **The Climber:** A "Victorian lift cage" aesthetic—brass, glass, and copper. It doesn't just "move"; it **ascends** with the grace of a specialized organ.
+
+---
+
+## 3. The Abyssal Axis (Undersea Elevator)
+
+A descent into the "Hidden Foundations." The pressure increases as we move toward the trench floor, and the architecture must **adapt or be mathematically false**.
+
+### Key Features
+- **The Shaft:** A rigid, bronze-riveted spine reaching 10,935m down.
+- **Research Stations:** Each station (1–5) is an "organelle" of the facility, its hull thickness and viewport size a direct function of the depth-pressure (Axiom: Form follows Depth).
+- **Sub Base:** A "deep-sea bunker" of brass and heavy rivets. The glass-floor observatory is the ultimate "The Living Axiom" test—looking directly into the void.
+
+---
+
+## 4. The Apex Outpost (Asteroid Spaceport)
+
+A 500m carbonaceous chondrite, "captured and nudged" to the counterweight position.
+
+### Key Features
+- **Hangars:** Carved directly into the rock, following the natural grain of the asteroid. Brass rolling doors and Amber guide lights provide the "Symmetry of Utility" for research vessels.
+- **The Spine:** The central bore where the "tether roots" terminate.
+- **Observation Dome:** The "eye" of the facility, looking back at the Earth-organism from 100,000 km.
+
+---
+
+## 5. Engine Cartography (Kaiju Integration)
+
+The **Engine Cartographer** will handle the transition between these four distinct scales:
+1.  **Macro-Space:** Earth orbit and the 100,000 km tether span.
+2.  **Atmospheric/Surface:** The floating island and its immediate surroundings.
+3.  **Deep-Ocean:** The 11km vertical descent through the 5 labs.
+4.  **Micro-Interior:** The detailed hallway and lab-benches (Sprint 3 focus).
+
+**Implementation Priority:**
+- **Skybox/Background:** (High) Tether and island visible from the hallway viewports.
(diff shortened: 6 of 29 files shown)


JamesPagetButler:
# Red Team Review — PR #401: Phase 4e Kaiju Virtual Lab

## Traffic-Light Summary

| Area | Reviewer | Verdict | Key Finding |
|------|----------|---------|-------------|
| Physics | Sabine | AMBER | Velocity derivation has dimensional error; DecayConstant is dimensionless proxy, not physical; single-envelope sinc2 omitted in qbp.FraunhoferIntensity |
| Architecture | Grothendieck | GREEN | Clean separation: instrument interfaces, room abstraction, qbp package isolation. Minor coupling issues in DSRoom. |
| Implementation | Knuth | AMBER | Rejection sampling can loop indefinitely for narrow fringes; per-particle mesh+entity allocation; duplicate physics engines between bakeoff and lab modes |

---

## Sabine Hossenfelder (Physics)

### 1. Dimensional Error in Eta Calculation

In `lab_physics.go:156`:

```go
v := hbar / (mass * e.Wavelength * 1e10) // velocity from de Broglie (simplified)
```

The de Broglie relation is `lambda = h / (m * v)`, giving `v = h / (m * lambda)`. The `hbar` used here is h-bar (h/2pi), not h, and the mysterious `1e10` factor has no documented justification. Working through the units:

- `hbar = 1.0545718e-34 J*s`
- `mass = 9.1093837e-31 kg`
- `Wavelength` = in meters (e.g., 5.5e-12 for Tonomura)

So: `v = 1.05e-34 / (9.1e-31 * 5.5e-12 * 1e10) = 1.05e-34 / (5.0e-32) = 2.1e-3 m/s`

But the actual electron velocity for a 50 pm de Broglie wavelength should be:
`v = h / (m * lambda) = 6.626e-34 / (9.1e-31 * 5.5e-12) = 1.32e8 m/s` (relativistic regime!)

The factor 1e10 brings the velocity down by 10 orders of magnitude, and using hbar instead of h drops it another factor of 6.28. The resulting eKinetic is wildly wrong, which means eta0 is wrong for any given U1. Since U1 values in the presets are chosen to produce "interesting" eta values, this error is currently masked — the demo works visually — but the eta is not physically meaningful.

**Recommendation:** Either (a) fix the velocity calculation and recalibrate U1 values, or (b) document explicitly that eta0 is a phenomenological fit, not derived from first principles. Option (b) is acceptable for a Phase 4e simulation prototype; option (a) is required before Phase 5 publication.

### 2. DecayConstant is a Dimensionless Proxy

In `qbp.go:57`:
```go
func DecayConstant(u1, d float64) float64 {
    return u1 * d
}
```

U1 is in Joules and d is in meters, so kappa = U1 * d has units of J*m. But kappa is used as `exp(-kappa * L)` where L is in meters, giving `exp(-(J*m^2))` — dimensionally meaningless. The Lean proof likely treats these as dimensionless Real numbers, but the Go simulation assigns physical units to U1 (Joules) and d (meters) via the presets.

This is documented as "simplified" in both the Go and Lean code, but the unit mismatch should be called out more prominently. The values "work" only because U1 values in the presets (1e-6, 1e-3) were hand-tuned to produce visible effects.

**Recommendation:** Add a comment block explaining the dimensional status of kappa, or introduce a dimensionally consistent formulation (e.g., kappa = U1 / (hbar * v) * d which would be dimensionless).

### 3. Missing Single-Slit Envelope in qbp.FraunhoferIntensity

The `qbp.FraunhoferIntensity` function (package level) computes only cos-squared:
```go
func FraunhoferIntensity(i0, d, lambda, l, x float64) float64 {
    arg := math.Pi * d * x / (lambda * l)
    return i0 * math.Cos(arg) * math.Cos(arg)
}
```

But `LabPhysicsEngine.FraunhoferIntensity` correctly includes both cos-squared and sinc-squared. The qbp package function is missing the single-slit envelope. This means any code calling `qbp.FraunhoferIntensity` directly (e.g., future differential tests) would get wrong results for finite slit widths. Currently nothing external calls it (the Lab physics engine uses its own method), but the asymmetry between the two implementations is a latent bug.

### 4. QBP Presets All Use Electron Mass

All presets use `mass := 9.1093837e-31` (electron mass) in the Eta() calculation, including `zeilinger_1988` which is a **neutron** experiment (lambda = 1.845 nm). The neutron mass is ~1836x larger, which would dramatically change eKinetic and thus eta0. This is another instance of the "phenomenological fit masks the physics" issue from point 1.

### 5. What IS Correct

- The Fraunhofer cos-squared * sinc-squared formula in LabPhysicsEngine is correct
- The QBP modification I(x) = (1-eta)*I_fraunhofer + eta correctly implements Model A
- The V = 1 - eta relationship is correctly wired
- Fringe spacing lambda*L/d is correct
- The Visibility function (Imax-Imin)/(Imax+Imin) is the standard definition
- Preset physical parameters (slit separations, wavelengths, screen distances) match the cited experiments

**Overall Physics Verdict: AMBER.** The simulation correctly demonstrates the QBP Model A phenomenology and the standard Fraunhofer physics. The dimensional issues in the eta derivation are real but do not affect the current use case (V&V of the simulation framework). They MUST be resolved before any quantitative predictions are published.

---

## Alexander Grothendieck (Architecture)

### 1. Clean Instrument Abstraction (POSITIVE)

The Device / Movable / Triggerable interfaces in `instrument.go` are well-designed. They mirror real lab instrument control frameworks (Bluesky/ophyd, TANGO) and provide clean boundaries:

- DSEmitter implements Movable (rate parameter)
- DSSlit implements Movable (separation, width)
- DSDetector implements Triggerable (start/stop, event stream)
- DSOracle implements Triggerable (predictions)

This is the right abstraction for a virtual physics lab.

### 2. qbp Package Isolation (POSITIVE)

The qbp package contains only pure math — no rendering, no Vulkan, no engine dependencies. The lab_physics.go file acts as a facade, delegating to qbp.* functions. This enables:
- Independent unit testing of physics
- Future FFI hot-swap to Lean/SciLean
- Differential testing without GPU

The verification chain (Lean proof -> Float oracle -> JSON -> Go tests) is architecturally sound.

### 3. Room Abstraction (POSITIVE)

ExperimentRoom interface + DSRoom implementation gives a clean path for future experiments. The LabGame delegates to ExperimentRoom.Update() and ExperimentRoom.Setup(), keeping the game loop experiment-agnostic.

### 4. DSRoom is Becoming a God Object (WARNING)

DSRoom at 700 lines owns: physics engine, coordinate mapper, 4 device instances, hit entities, desk controls, monitor text system, monitor content, UI manager, crosshair panels, help panel, audio system, oracle overlay entities, and V&V checklist. It has 15+ fields.

While each sub-system is well-encapsulated (DeskControlManager, MonitorTextSystem, VVChecklist), the room itself is accumulating orchestration responsibility. At this size it is manageable, but adding more features (GPU particles #397, hallway integration #393) will push it past the maintainability threshold.

**Recommendation:** Consider extracting a LabHUD struct (crosshair, help strip) and a ParticleRenderer struct (hitEntities, hitMesh, hitMat, spawning, cap cycling) to keep DSRoom under 500 lines.

### 5. Duplicate Physics Engines

DoubleslitGame (the original bakeoff prototype, doubleslit_game.go) has its own dsPhysicsEngine with duplicate Fraunhofer formulas. LabGame uses LabPhysicsEngine which delegates to qbp.*. The bakeoff engine does NOT include QBP (no eta, no visibility reduction). This means:

- Two copies of intensity calculations to maintain
- The bakeoff prototype silently diverges from the lab physics
- Bug fixes to one will not propagate to the other

**Recommendation:** Either (a) delete the bakeoff dsPhysicsEngine and make DoubleslitGame use LabPhysicsEngine with U1=0, or (b) add a deprecation comment to doubleslit_game.go noting it is the Phase 4e prototype, now superseded by lab_game.go.

### 6. Memory Management — Entity Lifecycle

The spawnHitDot function creates a new engine.Entity per particle but only calls entity.SetActive(false) on reset — never destroys them. With a 25K particle cap, this means up to 25K entities in the scene graph at all times. The hitMesh and hitMat are correctly shared (one allocation), but entity + transform allocations still accumulate.

For a Vulkan draw-call-bound renderer, 25K entities is significant. The current cap system prevents unbounded growth, but a pooled approach (pre-allocate N entities, cycle through them) would be more cache-friendly.

**Recommendation:** Acceptable for Phase 4e prototype. Flag for #397 (GPU particle optimization).

**Overall Architecture Verdict: GREEN.** The separation of concerns is strong. The instrument interfaces, qbp package isolation, and room abstraction are well-designed. The God object tendency in DSRoom and the duplicate physics engine are minor issues that should be tracked but do not block merge.

---

## Donald Knuth (Implementation)

### 1. Rejection Sampling Can Loop Indefinitely (BUG)

`LabPhysicsEngine.SampleHitPosition()` (line 210):
```go
func (e *LabPhysicsEngine) SampleHitPosition() float64 {
    halfWidth := 5.0 * e.FringeSpacing()
    for {
        x := (e.rng.Float64()*2 - 1) * halfWidth
        if e.rng.Float64() < e.Intensity(x) {
            return x
        }
    }
}
```

This is rejection sampling with an accept probability proportional to Intensity(x). The maximum intensity is 1.0 (at x=0), so the envelope is tight. However, for very narrow fringes (large d/lambda), most of the sampling window has near-zero intensity (the fringes are very sharp). The acceptance rate approaches 1/(number_of_fringes), which for pathological parameters could be extremely low.

The same pattern exists in dsPhysicsEngine.SampleHitPosition() and UISmokeTestGame.update().

**Recommendation:** Add a loop counter with a safety cap (e.g., 10000 iterations) and return the last candidate if the cap is hit. This prevents frame stalls without materially affecting the distribution.

### 2. Per-Particle Entity + Drawing Allocation (PERFORMANCE)

DSRoom.spawnHitDot() creates a new engine.Entity for every particle hit. At the default 500 Hz emission rate, this is 500 heap allocations per second for entities + 500 drawing registrations. The mesh and material are correctly shared, but:

- engine.NewEntity(host.WorkGroup()) allocates
- host.Drawings.AddDrawing(drawing) appends to a slice (potential reallocation)
- Each entity Transform is independently updated

The hitEntities slice in DSRoom is NOT pre-allocated — it grows via append.

**Recommendation:** Pre-allocate hitEntities to maxParticles capacity: `r.hitEntities = make([]*engine.Entity, 0, r.maxParticles)`. Also consider an entity pool for recycling when the cap is hit, rather than SetActive(false) + new allocation.

### 3. Oracle Overlay: 100 Spheres Per Rebuild

buildOracleOverlay creates 100 sphere meshes. rebuildOracleOverlay updates their positions but does not reuse the mesh allocation. The mesh NewMeshSphere(..., 0.008, 3, 6) is created per-sphere despite being identical — it should use the mesh cache key to share a single mesh (like hitMesh does for particles).

### 4. Histogram Bin-to-Bar Mapping: Integer Division Truncation

`lab_monitors.go:138`:
```go
binsPerBar := dsHistogramBins / monitorHistBars  // 100 / 40 = 2 (truncated)
```

100 bins / 40 bars = 2.5 bins per bar, but integer division gives 2. This means only 80 of the 100 histogram bins are displayed; the last 20 bins (the rightmost 20% of the pattern) are silently dropped. For a symmetric Fraunhofer pattern, this clips the outer fringes.

**Recommendation:** Use floating-point bin mapping or adjust monitorHistBars to evenly divide dsHistogramBins (e.g., 50 bars for 100 bins = 2 bins/bar exactly, or 25 bars = 4 bins/bar).

### 5. Watchdog Timer: 5-Second Timeout + os.Exit(1)

The heartbeat watchdog in both LabGame and DoubleslitGame exits the process after 5 seconds of update loop stall. This is a sensible GPU-hang recovery mechanism, BUT:

- The goroutine leaks the heartbeat channel (never closed on normal exit)
- os.Exit(1) skips all defer cleanup (including fpsLog.Close())
- The SIGUSR1 handler similarly uses os.Exit(1)

For a prototype, this is acceptable. For production, use context.WithCancel and graceful shutdown.

### 6. FPS Log File: Unbuffered Sync Per Second

`lab_game.go:549`:
```go
g.fpsLog.Sync()
```

Sync() forces an fsync, which is unnecessarily heavy for a debug log. Just letting the OS buffer handle it would be sufficient.

### 7. Visibility Measurement: Edge-Bin Exclusion

`ds_devices.go:226`:
```go
for i := 5; i < dsHistogramBins-5; i++ {
```

Excluding the first and last 5 bins (5% of each edge) is a reasonable heuristic to avoid boundary effects, but the choice of 5 is arbitrary and not documented. For wide fringe patterns (few fringes, e.g., Bach 2013), 5 bins might exclude relevant data.

**Overall Implementation Verdict: AMBER.** The code is well-structured and mostly correct. The rejection sampling infinite loop risk and the histogram truncation bug are the two items that should be fixed before merge. The per-particle allocation pattern is acceptable for Phase 4e but should be flagged for #397.

---

## Acceptance Criteria Verification

**Issue:** #302 — Experiment 03: Phase 4e — Verified Simulation Engine

### AC1: "Go simulation scene implemented for Experiment 03"

**Verdict: PASS**

Evidence: lab_game.go implements a full 3D lab room with optical bench, barrier, detector, source, desk, monitors, and chair. ds_room.go implements the ExperimentRoom interface for Experiment 03. The scene includes: floor with grid, walls, ceiling, raised platform, desk with monitors, chair, optical bench with apparatus. Build and run instructions are documented in README.md.

### AC2: "Physics oracle provides Lean-proven predictions"

**Verdict: PARTIAL**

Evidence: ds_devices.go:DSOracle loads predictions from tests/oracle_predictions.json. The qbp package functions are documented as mirroring FloatCompute.lean with theorem citations. FraunhoferIntensity, Visibility, FringeSpacing, QuatFraction, and CouplingDecomposition all cite specific Lean theorems. However:
- The oracle file loading uses 4 hardcoded relative paths (fragile)
- The qbp.FraunhoferIntensity omits the sinc-squared envelope (see Sabine finding #3)
- The Eta() calculation has dimensional issues (see Sabine finding #1)
The oracle pipeline works, but the physics fidelity has documented gaps.

### AC3: "Human can manipulate experiment setup parameters"

**Verdict: PASS**

Evidence: desk_controls.go:DeskControlManager provides 12+ interactive controls on the desk surface: Start/Stop, Rate Up/Down, Reset, Oracle Toggle, Slit Dial, Cap Cycle, and 5 preset buttons (Bach, Zeilinger, Tonomura, QBP weak, QBP strong). ds_room.go:HandleInput maps keyboard shortcuts (5-9 for presets, +/- for rate, O for oracle, V for V&V, R for reset, Enter for emitter, Space for freeze). The FPS camera supports WASD movement, mouse look, sit/stand, and 4 preset views.

### AC4: "V&V checklist documented and completed"

**Verdict: PARTIAL**

Evidence: docs/verification/vv_checklist_03.md documents a 13-item checklist covering Setup Verification, Behavior Verification, and Results Validation. vv_checklist.go implements an in-game V&V panel with 3 auto-checked criteria (N>1000, fringe spacing <5%, visibility) and 2 manual criteria (pattern match, no anomalies). However, the document Actual and Pass columns are all empty — the checklist has been documented but not yet completed (requires James visual verification run). The sign-off table has no signatures.

### Requirements Checklist

| Requirement | Status | Evidence |
|-------------|--------|----------|
| Implement experiment-specific Go scene for Double-Slit | PASS | 13 Go files, 4485 LOC |
| Load oracle predictions from Phase 4d (#301) | PASS | DSOracle.loadPredictions() loads oracle_predictions.json |
| V&V workflow: setup -> behavior -> results | PASS | In-game checklist + documented checklist |
| Document V&V checklist for Experiment 03 | PARTIAL | Documented but not completed (needs James) |

### Overall AC Verdict: PARTIAL

AC1 and AC3 are clearly PASS. AC2 is PARTIAL due to the dimensional issues in the eta derivation (the oracle pipeline works, but physics fidelity has gaps). AC4 is PARTIAL because the checklist exists but has not been walked through and signed off by the human reviewer. Both PARTIAL items are expected to be resolved during James visual verification pass — they are not architectural or design issues.

---

## Summary of Blocking / Non-Blocking Issues

| # | Issue | Severity | Recommendation |
|---|-------|----------|----------------|
| S1 | Dimensional error in velocity/eta derivation | AMBER | Fix or document as phenomenological; required before Phase 5 |
| S2 | DecayConstant dimensional mismatch | AMBER | Document the unit status explicitly |
| S3 | qbp.FraunhoferIntensity missing sinc-squared | AMBER | Add sinc-squared term to match lab physics engine |
| S4 | Hardcoded electron mass for all presets | AMBER | Use particle mass from preset or document limitation |
| A1 | DSRoom accumulating fields (proto-God-object) | LOW | Extract LabHUD + ParticleRenderer before hallway integration |
| A2 | Duplicate physics engine in bakeoff prototype | LOW | Deprecate or delete dsPhysicsEngine |
| K1 | Rejection sampling infinite loop risk | AMBER | Add iteration cap (10K) with fallback |
| K2 | Histogram bar truncation (80/100 bins shown) | AMBER | Fix integer division or adjust bar count |
| K3 | Per-particle entity allocation (no pool) | LOW | Flag for #397; pre-allocate hitEntities capacity now |
| K4 | Oracle overlay: 100 non-shared sphere meshes | LOW | Use mesh cache key for sharing |

**No BLOCKING issues found.** All AMBER items are real but acceptable for a Phase 4e prototype. The physics dimensional issues (S1-S4) should be tracked as a housekeeping issue and resolved before Phase 5 publication.

---

*Red Team Review by Sabine Hossenfelder, Alexander Grothendieck, and Donald Knuth*
*PR #401 — Phase 4e: Kaiju Virtual Lab with Scientist UX*
*Reviewed: 2026-02-22*

JamesPagetButler:
# Gemini Review — PR #401: Phase 4e Kaiju Virtual Lab

**Reviewers:** Cohl Furey (Mathematical Physics), Richard Feynman (Practical Physics)
**Model:** gemini-3-pro-preview (with thinking)

---

## Traffic-Light Summary

| Area | Reviewer | Verdict | Key Finding |
|------|----------|---------|-------------|
| Dimensional Analysis | Furey | RED | `Eta()` uses hbar instead of h, 1e10 factor is incorrect unit conversion, velocity off by ~20 orders of magnitude |
| Experimental Integrity | Feynman | RED | Zeilinger preset uses electron mass for neutron experiment (1836× error) |
| Rejection Sampling Safety | Feynman | RED | Infinite loop risk in `spawnParticleBatch` — no max-iteration guard |
| Model A Implementation | Furey | GREEN | I = (1-η)I_QM + η*I_Noise correctly structured |
| Fraunhofer Intensity | Feynman | GREEN | cos² × sinc² correctly implemented in lab version |
| V&V Checklist | Feynman | AMBER | Missing "null hypothesis" check (eta=0 baseline) |
| DecayConstant Units | Furey | AMBER | kappa has J·m units, should be m⁻¹ for dimensionless exponent |
| U1 Naming | Furey | AMBER | Variable name doesn't convey physical meaning |

## Detailed Findings

### RED: Dimensional Error in `Eta()` (Furey)

The formula `v := hbar / (mass * e.Wavelength * 1e10)` is dimensionally incorrect:
- Uses ℏ (h-bar) instead of h (Planck's constant) — introduces 2π factor error
- The `1e10` in the denominator: if Wavelength is in Angstroms, conversion to meters is `× 1e-10`. Placing `1e10` in the denominator creates a 10²⁰ magnitude error
- Result: calculated velocity is ~20 orders of magnitude wrong
- Consequence: eta0 is physically meaningless, though masked by hand-tuned U1 values

### RED: Wrong Particle Mass (Feynman)

All presets hardcode `mass := 9.1093837e-31` (electron). Zeilinger 1988 used **neutrons** (1836× heavier). This completely falsifies fringe spacing and visibility predictions for that preset. A physicist would immediately reject the tool.

### RED: Infinite Loop Risk (Feynman)

`spawnParticleBatch` has `for { ... }` rejection sampling with no exit condition. If `maxI` is high but intensity is low everywhere (parameter error, destructive interference), the render thread hangs permanently.

**Fix:** Add max-iteration counter (e.g., 100 tries), force-break with warning.

### AMBER: DecayConstant Units (Furey)

`qbp.DecayConstant(u1, d)` returns u1*d with units J·m. Used in `math.Exp(-kappa * ScreenDist)` where ScreenDist is meters → exponent has units J·m² — dimensionally undefined. Preset U1 values are hand-tuned to compensate.

### AMBER: Missing Null Hypothesis Check (Feynman)

V&V checklist should include a baseline check: set eta=0 and confirm simulation matches standard QM exactly. This is the first thing any skeptic will try.

### GREEN: Model A Structure (Furey)

The architecture correctly isolates the noise term (eta) from standard intensity: `I = (1-η)*I_QM + η*I_Noise`. This adheres to Model A structure.

### GREEN: Fraunhofer Intensity (Feynman)

cos² × sinc² correctly implemented in `LabPhysicsEngine.FraunhoferIntensity`. Standard wave optics.

## Acceptance Criteria Verification

| AC | Criterion | Verdict | Evidence |
|----|-----------|---------|----------|
| AC1 | Go scene with Vulkan rendering | **PASS** | 15 Go source files, Kaiju engine, Vulkan pipeline |
| AC2 | Physics oracle with QBP predictions | **PARTIAL** | Oracle exists but dimensional errors in eta calculation |
| AC3 | Interactive parameter manipulation | **PASS** | NASA-style desk controls with preset switching |
| AC4 | V&V checklist for experiment verification | **PARTIAL** | Checklist exists but missing null-hypothesis baseline |

## Verdict: **NOT READY FOR MERGE**

The architecture and visual implementation are solid, but the physics engine contains critical errors that invalidate the scientific utility of the simulation. The dimensional analysis must be corrected before merging.

**Feynman:** "It's a beautiful lab, but the machinery inside is broken. If the mass is wrong and the velocity is off by 20 orders of magnitude, we aren't doing physics; we're just making pretty pictures."

**Furey:** "The translation from the Lean proofs to Go code has failed at the unit analysis level. The equations for Model A are present but fed with garbage data."

### Must-Fix Items
1. Fix `h` vs `hbar` and unit conversion in velocity calculation
2. Set particle mass per-preset (electron vs neutron)
3. Add max-iteration guard to rejection sampling loop
4. Verify DecayConstant dimensional consistency

### Should-Fix Items
5. Add "Standard QM" (eta=0) baseline to V&V checklist
6. Document U1 naming convention

JamesPagetButler:
# Issues Identified — PR #401 Synthesis

Both Red Team (Sabine, Grothendieck, Knuth) and Gemini (Furey, Feynman) reviews complete. Here is the combined assessment.

## Combined Traffic-Light Table

| # | Issue | Red Team | Gemini | Combined | Blocking? |
|---|-------|----------|--------|----------|-----------|
| S1 | `Eta()` dimensional error (hbar vs h, 1e10 factor) | AMBER (Sabine) | RED (Furey) | **RED** | Yes — physics invalid |
| S2 | Electron mass hardcoded for all presets (neutron experiment wrong) | AMBER (Sabine) | RED (Feynman) | **RED** | Yes — experimental fraud risk |
| S3 | Rejection sampling infinite loop | AMBER (Knuth) | RED (Feynman) | **RED** | Yes — render thread can hang |
| S4 | DecayConstant kappa units (J·m not m⁻¹) | AMBER (Sabine) | AMBER (Furey) | **AMBER** | No — masked by tuned presets |
| S5 | qbp.FraunhoferIntensity missing sinc² | AMBER (Sabine) | — | **AMBER** | No — lab version correct |
| S6 | Missing null-hypothesis V&V check (eta=0 baseline) | — | AMBER (Feynman) | **AMBER** | No — UX improvement |
| S7 | DSRoom trending God-object (700 lines) | GREEN (Grothendieck) | — | **GREEN** | No |
| S8 | Per-particle entity allocation (no pooling) | AMBER (Knuth) | — | **AMBER** | No — #397 will address |
| S9 | Histogram shows 80/100 bins | AMBER (Knuth) | — | **AMBER** | No — cosmetic |

## Agreement/Disagreement

Both teams **agree** on all findings. No conflicts. Gemini escalated Sabine's AMBER items S1-S3 to RED, which is appropriate — the dimensional errors are critical for scientific credibility even in a prototype.

## Proposed Fix Plan

**Fix in this PR (RED items):**
1. **S1:** Fix velocity: use `h` (not hbar), correct wavelength conversion
2. **S2:** Add per-preset particle mass lookup (electron, neutron)
3. **S3:** Add max-iteration guard (100 attempts) to rejection sampling

**Defer (AMBER items):**
- S4: DecayConstant units → housekeeping issue (requires Lean coordination)
- S5: qbp.FraunhoferIntensity sinc² → housekeeping issue
- S6: Null-hypothesis V&V check → can add in follow-up
- S8: Entity pooling → #397
- S9: Histogram bins → cosmetic fix

## @JamesPagetButler — Decision Needed

Fix S1-S3 in this PR, or defer all to housekeeping issues?

JamesPagetButler:
## S1-S3 Fixes Applied (commit eba45d8)

### S1: Eta() dimensional error — FIXED
- **Was:** `hbar / (mass * wavelength * 1e10)` — used ℏ instead of h, mystery `1e10` factor
- **Now:** `h / (e.Mass * e.Wavelength)` — correct de Broglie: v = h/(mλ)
- Planck's constant: `6.62607015e-34` J·s (2019 SI exact value)
- Recalibrated QBP U₁ values against corrected E_kinetic ≈ 8.0e-15 J:
  - **qbp_weak:** 1e-6 → 2.5e-15 J (η₀≈0.09, V≈0.91)
  - **qbp_strong:** 1e-3 → 1.2e-14 J (η₀≈0.69, V≈0.31)

### S2: Per-preset particle mass — FIXED
- Added `Mass float64` to `Preset` struct and `LabPhysicsEngine`
- Bach, Tonomura, QBP presets: `electronMass = 9.1094e-31 kg`
- Zeilinger: `neutronMass = 1.6749e-27 kg`
- `ApplyPreset()` now carries mass

### S3: Rejection sampling guard — FIXED
- `SampleHitPosition()`: `for { ... }` → `for i := 0; i < 10_000; i++ { ... }`
- Returns x=0 (center of pattern) on exhaustion
- Prevents infinite loop with degenerate parameters

### Verification
- ✅ Build passes (Go 1.25.0)
- ✅ All qbp package tests pass (10/10)
- ✅ engine/src/ copy synced

JamesPagetButler:
## Red Team Review — Round 2 (Post S1-S3 Fix)

**Reviewer:** Claude (Hossenfelder / Grothendieck / Knuth personas)
**Commit under review:** eba45d8 ("fix(physics): Correct Eta() dimensional errors, add per-preset mass, guard rejection sampling")
**Scope:** Physics correctness of S1-S3 fixes in `lab_physics.go` and `qbp/qbp.go`.

---

### Sabine Hossenfelder — Physics Rigor

**Verdict: AMBER**

**S1 — de Broglie formula: Correct but non-relativistic.**

The fix is dimensionally correct: `v = h/(m*lambda)`, `E_k = 0.5*m*v^2`. Planck's constant `h = 6.62607015e-34` is the 2019 SI exact definition. Good.

However, the non-relativistic approximation introduces significant error for two presets:

| Preset | v/c | gamma | E_k error vs. relativistic |
|--------|-----|-------|---------------------------|
| Bach 2013 (50 pm) | 0.049 | 1.001 | 0.2% — acceptable |
| Tonomura 1989 (5.5 pm) | 0.441 | 1.114 | **14.9%** — significant |
| Zeilinger 1988 (1.845 nm) | 7e-7 | 1.000 | 0.0% — negligible |
| QBP presets (5.5 pm) | 0.441 | 1.114 | **14.9%** — inherited from Tonomura |

For Tonomura electrons at 0.44c, the relativistic kinetic energy is `E_k = (gamma-1)*mc^2 = 9.36e-15 J` vs. the non-relativistic `0.5*m*v^2 = 7.97e-15 J`. The 15% error propagates into the eta0 calculation. This is acceptable for a visualization engine with a documented limitation, but the comment claiming `E_kinetic ~ 8.0e-15 J` should be annotated as the non-relativistic value. **AMBER — document-and-defer, not blocking.**

**S2 — Per-preset mass: Correct.**

`electronMass = 9.1093837015e-31 kg` matches 2018 CODATA. `neutronMass = 1.67492749804e-27 kg` matches 2018 CODATA. Zeilinger uses neutrons; the rest use electrons. All correct.

**U1 Recalibration: Values are self-consistent but dimensionally suspect.**

The eta0 formula `U1^2 / (E_k^2 + U1^2)` is dimensionally homogeneous: both terms in the denominator are J^2, eta0 is dimensionless. The computed values match the claimed targets:
- qbp_weak: U1=2.5e-15 J => eta0=0.090, V=0.91 (claimed ~0.09, ~0.91)
- qbp_strong: U1=1.2e-14 J => eta0=0.694, V=0.31 (claimed ~0.69, ~0.31)

**DecayConstant dimensional inconsistency — the sleeping issue.**

The Lean proof (DoubleSlit.lean line 417-421) explicitly documents:
> "Dimensionless coupling proxy: product of potential strength and slit separation. This is a simplified stand-in used for monotonicity and positivity proofs. The physical decay constant kappa = |U1|/(hbar*v) [m^-1] from the ground truth has different dimensions."

The Go code uses the Lean proxy (`kappa = U1 * d`) instead of the physical formula (`kappa = |U1|/(hbar*v)`). Dimensionally:
- Code: `kappa = J * m`, then `exp(-kappa * L)` = `exp(-J*m^2)` — **not dimensionless**
- Physical: `kappa = J/(J*s * m/s) = m^-1`, then `exp(-kappa * L)` = `exp(-dimensionless)` — correct

**Numerically, it does not matter for these presets** because:
- Code kappa_weak * L = 2.5e-21 => exp(-2.5e-21) = 1.0000000000 (no decay)
- Code kappa_strong * L = 1.2e-20 => exp(-1.2e-20) = 1.0000000000 (no decay)

So the decay term is effectively 1.0, and eta = eta0 for all current presets. The entire decay mechanism is a no-op. This means:
1. The DecayConstant function could be replaced with `return 0` for identical behavior.
2. If someone adds presets with larger U1 or d values, the non-physical kappa will produce garbage results.
3. The Lean proof documents that the *physical* formula is `kappa = |U1|/(hbar*v)`, but the Go code ignores this.

**Severity: AMBER (non-blocking) for Phase 4e.** The visualization works correctly because the decay is negligible. But this should be tracked as a Phase 2 issue: when the Python simulation uses the physical formula, the Go visualization engine should match. **Recommend creating a housekeeping issue.**

---

### Alexander Grothendieck — Mathematical Structure

**Verdict: GREEN**

**eta0 bounds: Proven correct.**

`eta0 = U1^2 / (E_k^2 + U1^2)` is of the form `x^2 / (y^2 + x^2)` where x,y >= 0. This maps to [0, 1]:
- When U1 = 0: eta0 = 0 (standard QM, matches `scenarioA_visibility`)
- When U1 >> E_k: eta0 -> 1 (which-path, matches `scenarioB_visibility`)
- The function is monotonically increasing in |U1|/E_k

The `if eta > 1 { eta = 1 }` guard is structurally correct but mathematically redundant: since eta0 in [0,1] and exp(-kappa*L) in (0,1], their product is in [0,1). This guard is harmless defensive code against floating-point edge cases. Acceptable.

**Intensity bounds: Correct.**

`I(x) = (1-eta)*I_fraunhofer(x) + eta` where eta in [0,1] and I_fraunhofer in [0,1].
- Maximum: at x=0, I_fraunhofer=1, so I(0) = (1-eta)*1 + eta = 1. Always 1.
- Minimum: at fringe minima where I_fraunhofer=0, I(x) = eta >= 0.
- Range: [eta, 1]. For standard QM (eta=0): [0,1]. For full coupling (eta=1): [1,1].

This guarantees rejection sampling is well-defined (max = 1 as the envelope).

**Rejection sampling convergence.**

For eta > 0: I(x) >= eta > 0 everywhere, so acceptance probability >= eta. Expected iterations: at most 1/eta. For qbp_weak (eta=0.09): ~11 iterations. For qbp_strong (eta=0.69): ~1.4 iterations. Well within 10,000.

For eta = 0 (standard QM): I(x) = cos^2 * sinc^2, which has zeros. But the sampling domain is [-5*fringe_spacing, 5*fringe_spacing], containing ~10 bright fringes. The average I(x) over this domain is approximately 0.25-0.5 (depending on slit width ratio), so expected iterations ~2-4.

The only degenerate case is eta=0 with a/d ratio causing the single-slit envelope to suppress all fringes within the sampling window. This is physically unreasonable for the given presets and would require pathological parameter combinations.

**Structure of the decay chain: Algebraically sound but physically dormant.**

eta = eta0 * exp(-kappa*L). The algebraic properties (monotonicity, positivity, limit behavior) all hold regardless of the numerical value of kappa. The Lean proofs for `decayConstant_pos` and `decayConstant_mono_U1` apply to the proxy formula. The issue is physical interpretation, not mathematical structure.

---

### Donald Knuth — Code Quality

**Verdict: GREEN**

**Constants accuracy.**

| Constant | Code value | CODATA 2018 | Match? |
|----------|-----------|-------------|--------|
| electronMass | 9.1093837015e-31 | 9.1093837015(28)e-31 | Exact central value |
| neutronMass | 1.67492749804e-27 | 1.67492749804(95)e-27 | Exact central value |
| h (Planck) | 6.62607015e-34 | 6.62607015e-34 (exact, 2019 SI) | Exact |

All three constants are correct.

**10,000-iteration guard: Appropriate.**

For I(x) with max value 1, the acceptance probability per trial is the average intensity over the sampling window. Even in the worst realistic case (standard QM, narrow fringes), the average is > 0.01. P(10000 consecutive rejections) < (0.99)^10000 ~ 2e-44. The guard will never fire under physically reasonable parameters.

**Fallback value (x=0): Acceptable.**

Returning 0 (center of pattern) biases toward the central maximum. For a degenerate case that exhausts 10,000 iterations, the parameters are already non-physical, so any fallback is equally arbitrary. Center is a defensible choice — it is the mode of all physically reasonable double-slit distributions.

**Code organization.**

- `Mass` field correctly added to both `Preset` struct and `LabPhysicsEngine`
- `ApplyPreset` copies all 6 fields including `Mass` — no field missed
- `NewLabPhysicsEngine()` defaults to `presets[2]` (Tonomura) which sets Mass to electronMass
- No magic numbers remaining — the mystery `1e10` from S1 is gone
- Comments accurately describe the physics intent and link to Lean proofs
- The `h` variable is local to `Eta()`, not a package constant — minor style note, but acceptable since it is only used in one function

**Minor suggestion (non-blocking):** The comment on line 77 says `E_kinetic ~ 8.0e-15 J` which is the non-relativistic value. Since the presets inherit Tonomura parameters where the relativistic E_k is 9.4e-15 J (15% higher), consider annotating this as `E_kinetic_NR ~ 8.0e-15 J`. Trivial fix, not blocking.

---

### Overall Verdict

| Persona | Verdict | Blocking? |
|---------|---------|-----------|
| Hossenfelder (Physics) | AMBER | No |
| Grothendieck (Math) | GREEN | No |
| Knuth (Code) | GREEN | No |

**PASS — no blocking issues.** The S1-S3 fixes are correct and the code is safe for merge.

**Non-blocking items to track:**

1. **AMBER: Non-relativistic approximation** — 15% E_k error for Tonomura/QBP presets (v=0.44c). Document as known limitation. If the visualization ever claims quantitative agreement with Tonomura data, this needs relativistic kinematics (`E_k = (gamma-1)*mc^2`).

2. **AMBER: DecayConstant is a dimensionless proxy, not physical kappa.** The Lean proof (DoubleSlit.lean line 419) documents that the physical formula is `kappa = |U1|/(hbar*v) [m^-1]`. The Go code uses the Lean algebraic proxy `kappa = U1*d` which is dimensionally incorrect (`[J*m]`). Currently harmless because the resulting decay is ~exp(-1e-21) = 1.0 for all presets, making the decay mechanism a no-op. Recommend a housekeeping issue: when decay behavior matters (e.g., short screen distances or large U1), the physical formula must be used.

3. **TRIVIAL: Comment annotation.** The `E_kinetic ~ 8.0e-15 J` comment should note this is the non-relativistic value.

JamesPagetButler:
## Gemini Review — Round 2 (Furey + Feynman)

### Cohl Furey (Mathematical Physics): RED
The η₀ formula correctly mirrors Lean proofs. However, `DecayConstant = U₁·d` has units [J·m], making `exp(-κ·L)` dimensionally broken (exponent = J·m²). Even though the decay is currently a no-op (≈1.0), shipping dimensionally invalid math undermines the engine's theoretical foundation.

### Richard Feynman (Practical Physics): AMBER
Rejection sampling guard is safe. Per-preset masses are correct (CODATA 2018). Standard QM correctly produces η=0, V=1. The non-relativistic de Broglie at v≈0.44c introduces ~15% error in E_k. Cheap to fix with relativistic formula.

### Overall: RED (requires revisions)
Two issues flagged:
1. **DecayConstant dimensionality** — exponent must be dimensionless
2. **Relativistic E_k** — use `E_total = sqrt((pc)² + (mc²)²)`, `E_k = E_total - mc²`

S2 (per-preset mass) and S3 (rejection sampling guard) are excellent — keep exactly as-is.

JamesPagetButler:
## Review Synthesis — Round 2

### Traffic-Light Table

| ID | Issue | Red Team | Gemini | Verdict | Action |
|----|-------|----------|--------|---------|--------|
| R1 | DecayConstant dimensionality (κ=U₁·d has units J·m, exponent not dimensionless) | AMBER | RED | **CONFLICT** | See analysis below |
| R2 | Non-relativistic E_k at v≈0.44c (~15% error) | AMBER | AMBER | AMBER | Defer or fix |
| R3 | S1 fix (h vs ℏ, remove 1e10) | GREEN | GREEN | GREEN | Done ✓ |
| R4 | S2 fix (per-preset mass) | GREEN | GREEN | GREEN | Done ✓ |
| R5 | S3 fix (rejection sampling guard) | GREEN | GREEN | GREEN | Done ✓ |

### CONFLICT: R1 — DecayConstant Dimensionality

**Red Team (AMBER):** The Lean proofs explicitly document this as a simplified algebraic proxy (DoubleSlit.lean §8). The decay evaluates to exp(-1e-21) ≈ 1.0 for all presets — it's a mathematical no-op. Not blocking.

**Gemini (RED):** Exponents must be dimensionless. Shipping broken math undermines theoretical foundation.

**Claude's analysis:** Both are correct from their perspectives. The key context:
1. The Lean proofs use **abstract algebra** (no SI units). `DecayConstant = u1 * d` is proven correct in that system.
2. The Go code plugs SI values into these algebraic formulas. The dimensional inconsistency is **by design** — it mirrors the Lean proof structure.
3. Making κ physically correct (κ = |U₁|/(ℏ·v)) would give κ ≈ 1.8e11 m⁻¹, causing exp(-κ·L) → 0, destroying ALL quaternionic effects.
4. The correct long-term fix is a **physically-grounded decay model** (tracked in the Lean proofs as future work), not a quick SI patch.

**Recommendation:** AMBER (defer). Document the dimensional mismatch with a comment. Create a housekeeping issue to track the physical decay model.

### R2 — Relativistic E_k
At v≈0.44c, γ≈1.11. The ~15% E_k error affects η₀ calibration, but since U₁ was calibrated against the same non-relativistic formula, the system is self-consistent. Fixing it is cheap (4 lines) but requires U₁ recalibration.

**Recommendation:** Fix now (cheap, more correct) or defer (self-consistent as-is).

JamesPagetButler:
## Round 2 Fixes Applied (commit af8ab55)

### R2: Relativistic E_k — FIXED
- Replaced `E_k = ½mv²` with `E_total = sqrt((pc)² + (mc²)²), E_k = E_total - mc²`
- Tonomura electrons at v/c=0.441: correction is ~4.4% (not 15% as initially estimated — Bach is only v/c=0.049, nearly non-relativistic)
- Recalibrated U₁: weak 2.5e-15→2.4e-15, strong 1.2e-14→1.1e-14

### R1: DecayConstant dimensionality — DEFERRED
- Documented the mismatch in code comment (lines 184-189)
- Created housekeeping issue #402 to track proper physical decay model
- Current impact: zero (decay evaluates to 1.0 for all presets)

### Status
All RED and AMBER items from Round 1 and Round 2 are resolved or deferred with James's approval. Ready for final review or merge decision.

Issue #1, opened 2026-02-01
Establish Visualization Framework


JamesPagetButler:
# Issue #001: Establish Visualization Framework

**Labels:** `infrastructure`, `accessibility`
**Status:** Open

## Summary

Select and configure a visualization stack for quaternion physics animations. Must support accessibility (dyslexia-friendly: motion over dense text, clear labels, high contrast).

## Requirements

| Requirement | Priority |
|-------------|----------|
| 3D quaternion rotation visualization | High |
| Animation export (GIF/MP4/HTML) | High |
| Browser-viewable output | Medium |
| Reproducible from code (no GUI) | High |
| Clear, high-contrast labeling | High |

## Candidates

| Framework | Use Case | Install |
|-----------|----------|---------|
| **manim** | Publication-quality math animations | `pip install manim` |
| **vpython** | Interactive 3D physics | `pip install vpython` |
| **plotly** | Interactive browser plots | `pip install plotly` |

**Recommendation:** `manim` for paper figures + `vpython` for exploration.

## Acceptance Criteria

- [ ] Framework(s) documented in `docs/visualization.md`
- [ ] Dependencies added to `requirements.txt`
- [ ] Minimal example: `src/viz/quaternion_rotation.py` animates a unit quaternion
- [ ] Output viewable in browser

## Structure

```
src/
  viz/
    __init__.py
    quaternion_rotation.py   # minimal working example
docs/
  visualization.md           # usage guide
```

## Branch

`feature/visualization-framework`

PR #508, opened 2026-06-04
[infra] PR check-status notifier — machine-stamped check verdict per head SHA

JamesPagetButler:
## Summary

One workflow: when all check suites for a PR's head SHA complete, **upsert a single marked comment** with the verdict — ✅ `All checks green (N) @ timestamp` or ❌ failures with links. Edit-in-place (never comment spam); updates per push so it's always for the *current* head.

## Why (two incidents, same day)

1. A "ready to merge" claim was made without re-querying checks (the push had triggered new gates — red).
2. A red verdict was acted on after the gate had already re-run green.

Both are the same defect: humans and agents reading **stale check state**. This puts a machine-stamped, timestamped verdict in the PR conversation itself — the canonical "CI green" artifact the evidence-first rule (`docs/workflows/review_tiers.md`) wants linked.

## Mechanics

- Trigger: `check_suite: completed`, PR-attached suites only
- If any suite for the SHA still running → exits quietly (no partial verdicts)
- Failure = `failure | timed_out | cancelled | action_required`; neutral/skipped don't count
- Permissions: `pull-requests: write`, `checks: read` — default token, no secrets

## Self-test plan (the bootstrap)

- [ ] This PR's own checks complete → the notifier **cannot** report on this PR (the workflow only runs from the default branch for `check_suite` events) — so post-merge verification: next PR after merge gets the comment
- [ ] YAML validated (`yaml.safe_load` clean)
- [ ] CI green on this PR
- [ ] Post-merge: first comment appears on the next active PR; verify upsert (not spam) on a subsequent push

## CTH anchor impact
- [x] N/A — CI workflow only; no anchor-bearing paths.

### DIFF
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -60,21 +60,16 @@ jobs:
           NFAIL=$(echo "$FAILED" | jq 'length')
           TS=$(date -u +"%Y-%m-%d %H:%M UTC")
 
+          # NOTE: bodies are assembled with printf, not multi-line literals —
+          # YAML run-block indentation inside a quoted string becomes leading
+          # spaces that GitHub markdown renders as code blocks.
           if [ "$NFAIL" -eq 0 ]; then
-            BODY="$MARKER
-          ## ✅ All checks green — \`$TOTAL\` checks @ $TS
-
-          Head: \`$SHA\`. This comment is machine-posted and updates per head SHA; it is the canonical \"CI green\" evidence link for this PR (evidence-first rule, \`docs/workflows/review_tiers.md\`)."
+            BODY=$(printf '%s\n## ✅ All checks green — `%s` checks @ %s\n\nHead: `%s`. Machine-posted; updates per head SHA. Canonical "CI green" evidence link for this PR (evidence-first rule, `docs/workflows/review_tiers.md`).' \
+              "$MARKER" "$TOTAL" "$TS" "$SHA")
           else
             LIST=$(echo "$FAILED" | jq -r '.[] | "- ❌ [\(.name)](\(.html_url)) — \(.conclusion)"')
-            BODY="$MARKER
-          ## ❌ Checks FAILED — $NFAIL of $TOTAL @ $TS
-
-          Head: \`$SHA\`
-
-          $LIST
-
-          Machine-posted; updates per head SHA."
+            BODY=$(printf '%s\n## ❌ Checks FAILED — %s of %s @ %s\n\nHead: `%s`\n\n%s\n\nMachine-posted; updates per head SHA.' \
+              "$MARKER" "$NFAIL" "$TOTAL" "$TS" "$SHA" "$LIST")
           fi
 
           # Upsert: find existing marked comment, edit it; else create.
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -18,6 +18,11 @@
 
 name: PR check-status notifier
 
+# KNOWN LIMITATION: check_suite payloads carry an empty pull_requests[] for
+# FORK PRs, so fork PRs get no verdict comment (acceptable for this repo:
+# solo maintainer + agent sessions, same-repo branches only — documented per
+# review finding #3 so a future external PR's silence isn't a mystery).
+
 on:
   check_suite:
     types: [completed]
@@ -27,8 +32,18 @@ permissions:
   checks: read
   contents: read
 
+# collapse simultaneous firings for the same head SHA; keep only the latest
+concurrency:
+  group: pr-check-status-${{ github.event.check_suite.head_sha }}
+  cancel-in-progress: true
+
 jobs:
   notify:
+    # IMPORTANT: this job name is the notifier's own check-run name; the query
+    # below excludes it by name so the notifier never counts itself as pending
+    # (review finding #1 — without the self-filter, PENDING never reaches 0
+    # and the notifier structurally never posts).
+    name: pr-check-status-notify
     # only suites attached to a PR
     if: ${{ github.event.check_suite.pull_requests[0] != null }}
     runs-on: ubuntu-latest
@@ -43,9 +58,16 @@ jobs:
           set -euo pipefail
           MARKER="<!-- pr-check-status-notifier -->"
 
-          # All check runs for this SHA (across suites)
+          # All check runs for this SHA (across suites — per-SHA scoping is
+          # deliberate: the verdict covers ALL checks, not one suite), EXCLUDING
+          # the notifier's own run (self-filter, review finding #1).
+          # NOTE on pagination (review finding #8): `gh api --paginate --jq`
+          # emits one JSON array PER PAGE (concatenated docs), so we slurp and
+          # flatten to a single array before counting.
+          SELF="pr-check-status-notify"
           RUNS=$(gh api "repos/$REPO/commits/$SHA/check-runs?per_page=100" --paginate \
-                 --jq '[.check_runs[] | {name, status, conclusion, html_url}]')
+                 --jq '[.check_runs[] | {name, status, conclusion, html_url}]' \
+                 | jq -s 'add | map(select(.name != "'"$SELF"'"))')
 
           PENDING=$(echo "$RUNS" | jq '[.[] | select(.status != "completed")] | length')
           if [ "$PENDING" -gt 0 ]; then
@@ -73,8 +95,10 @@ jobs:
           fi
 
           # Upsert: find existing marked comment, edit it; else create.
+          # (slurp-flatten across pages, same pagination caveat as above)
           CID=$(gh api "repos/$REPO/issues/$PR/comments?per_page=100" --paginate \
-                --jq "[.[] | select(.body | startswith(\"$MARKER\"))] | (first // {}) | .id // empty")
+                --jq "[.[] | select(.body | startswith(\"$MARKER\"))]" \
+                | jq -s 'add | (first // {}) | .id // empty' -r)
           if [ -n "$CID" ]; then
             gh api -X PATCH "repos/$REPO/issues/comments/$CID" -f body="$BODY" > /dev/null
             echo "updated comment $CID"
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -1,0 +1,89 @@
+# PR check-status notifier — posts ONE upserted comment per PR with the final
+# check verdict for the current head SHA, timestamped.
+#
+# Why (2026-06-04, two incidents in one day): humans and agents kept acting on
+# STALE check states — a "ready to merge" claim made without re-querying, and a
+# red verdict read after the gate had already re-run green. This makes the PR
+# conversation itself carry a machine-stamped verdict for the current head, so
+# "CI green" claims have a canonical artifact to link (evidence-first rule,
+# docs/workflows/review_tiers.md).
+#
+# Behavior:
+#   - fires when any check suite completes
+#   - if other suites for the same SHA are still running -> exits quietly
+#   - else upserts (edit-in-place, never spam) a single marked comment:
+#       ALL GREEN (N checks) @ timestamp for sha
+#       or FAILED: <names + links> @ timestamp for sha
+#   - new push -> next completion updates the comment for the new SHA
+
+name: PR check-status notifier
+
+on:
+  check_suite:
+    types: [completed]
+
+permissions:
+  pull-requests: write
+  checks: read
+  contents: read
+
+jobs:
+  notify:
+    # only suites attached to a PR
+    if: ${{ github.event.check_suite.pull_requests[0] != null }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Compose and upsert verdict comment
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR: ${{ github.event.check_suite.pull_requests[0].number }}
+          SHA: ${{ github.event.check_suite.head_sha }}
+        run: |
+          set -euo pipefail
+          MARKER="<!-- pr-check-status-notifier -->"
+
+          # All check runs for this SHA (across suites)
+          RUNS=$(gh api "repos/$REPO/commits/$SHA/check-runs?per_page=100" --paginate \
+                 --jq '[.check_runs[] | {name, status, conclusion, html_url}]')
+
+          PENDING=$(echo "$RUNS" | jq '[.[] | select(.status != "completed")] | length')
+          if [ "$PENDING" -gt 0 ]; then
+            echo "$PENDING check(s) still running for $SHA — not posting yet."
+            exit 0
+          fi
+
+          TOTAL=$(echo "$RUNS" | jq 'length')
+          # neutral/skipped don't count as failures
+          FAILED=$(echo "$RUNS" | jq '[.[] | select(.conclusion as $c
+                    | ["failure","timed_out","cancelled","action_required"] | index($c))]')
+          NFAIL=$(echo "$FAILED" | jq 'length')
+          TS=$(date -u +"%Y-%m-%d %H:%M UTC")
+
+          if [ "$NFAIL" -eq 0 ]; then
+            BODY="$MARKER
+          ## ✅ All checks green — \`$TOTAL\` checks @ $TS
+
+          Head: \`$SHA\`. This comment is machine-posted and updates per head SHA; it is the canonical \"CI green\" evidence link for this PR (evidence-first rule, \`docs/workflows/review_tiers.md\`)."
+          else
+            LIST=$(echo "$FAILED" | jq -r '.[] | "- ❌ [\(.name)](\(.html_url)) — \(.conclusion)"')
+            BODY="$MARKER
+          ## ❌ Checks FAILED — $NFAIL of $TOTAL @ $TS
+
+          Head: \`$SHA\`
+
+          $LIST
+
+          Machine-posted; updates per head SHA."
+          fi
+
+          # Upsert: find existing marked comment, edit it; else create.
+          CID=$(gh api "repos/$REPO/issues/$PR/comments?per_page=100" --paginate \
+                --jq "[.[] | select(.body | startswith(\"$MARKER\"))] | (first // {}) | .id // empty")
+          if [ -n "$CID" ]; then
+            gh api -X PATCH "repos/$REPO/issues/comments/$CID" -f body="$BODY" > /dev/null
+            echo "updated comment $CID"
+          else
+            gh api -X POST "repos/$REPO/issues/$PR/comments" -f body="$BODY" > /dev/null
+            echo "created comment"
+          fi
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -32,9 +32,12 @@ permissions:
   checks: read
   contents: read
 
-# collapse simultaneous firings for the same head SHA; keep only the latest
+# collapse simultaneous firings per PR (not per SHA: PR-keying also cancels a
+# stale in-flight run for an OLD head SHA when a new push's suites complete,
+# so an outdated verdict can never overwrite a newer one — review concurrency
+# finding, round 1)
 concurrency:
-  group: pr-check-status-${{ github.event.check_suite.head_sha }}
+  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.check_suite.head_sha }}
   cancel-in-progress: true
 
 jobs:


JamesPagetButler:
## Red Team Review — PR #508 (PR check-status notifier)

**Persona:** Knuth (lead, correctness & semantics) + Sabine (sanity check)
**Tier:** 2 | **Scope:** single new file `.github/workflows/pr-check-status.yml` (84 lines)
**Method:** Static read + reasoning through GitHub API / Actions event semantics. Items I cannot confirm without merging to the default branch are marked **UNVERIFIED**.

### Verdict: 🔴 REQUEST CHANGES — one BLOCKING structural defect (self-counting, focus #3)

---

PR self-test plan (from body):

| Criterion | Status | Evidence |
|---|---|---|
| Notifier cannot report on this PR (runs from default branch for `check_suite`) | PASS (claim accurate) | `check_suite`/`check_run`/`workflow_run` always use the workflow file on the **default branch** — GitHub documented. The version on the PR branch is dormant until merged. |
| YAML validated (`yaml.safe_load` clean) | UNVERIFIED | Not re-run here; structurally well-formed. |
| CI green on this PR | UNVERIFIED | Defer to `gh pr checks`. |
| Post-merge: first comment on next PR, upsert not spam | **BLOCKING — predicted to FAIL** | See finding #1. |

---

### Findings

| # | Sev | Area | Finding |
|---|---|---|---|
| 1 | 🔴 BLOCKING | Self-counting (focus #3) | Notifier's own `notify` check run is counted as PENDING → never posts |
| 2 | 🟢 OK | Recursion (focus #2) | Bounded by GitHub's `GITHUB_TOKEN` recursion guard — no infinite loop |
| 3 | 🟡 NOTE | Fork PRs (focus #1) | `pull_requests[0]` empty for forks; acceptable for this repo, document it |
| 4 | 🟢 OK | Injection (focus #4) | No command injection; markdown injection benign |
| 5 | 🟡 NOTE | Counting semantics | `commits/{sha}/check-runs` is per-SHA, NOT per-suite — pulls in unrelated suites incl. the notifier itself |
| 6 | 🟢 OK | printf body (focus #6) | Leading-space code-block bug resolved by commit f43265b |
| 7 | 🟢 OK | Permissions | Minimal and correct |
| 8 | 🟡 NOTE | Pagination / upsert | Correct, with one race caveat |

---

### 🔴 Finding #1 — BLOCKING: the notifier counts its own run as pending → PENDING never reaches 0 → never posts

This is the defect class focus #3 flagged, and the timeline confirms it kills the design as written.

**The query is per-SHA, not per-suite:**
```
gh api repos/$REPO/commits/$SHA/check-runs?per_page=100
```
`commits/{sha}/check-runs` returns **every check run associated with that commit, across all check suites** — not the runs of the suite that fired the event. (The event payload's `check_suite` is scoped; this REST endpoint is not.)

**The notifier is itself a check run under the same SHA.** When a GitHub Actions workflow runs in response to an event for head SHA `S`, that workflow run produces its own check_suite and a check run (job name `notify`) **attached to SHA `S`**. So while the notifier job is executing its `gh api ... commits/$SHA/check-runs` call, its own `notify` run is in that list with `status: in_progress`.

**Timeline:**
1. Real suite (e.g. `lint`) completes → `check_suite:completed` fires → dispatches notifier run on SHA `S`.
2. Notifier's own `notify` check run is created under `S`, `status: in_progress`.
3. Step queries `commits/S/check-runs` → list includes the in-progress `notify` run.
4. `PENDING = [select(.status != "completed")] | length` ≥ 1 (itself).
5. `[ "$PENDING" -gt 0 ]` → "still running" → `exit 0`. **No comment, ever.**

Every firing spawns a notifier run that observes *itself* as pending at query time. There is no firing on which `PENDING` can be 0, because the act of querying is performed by a still-running check. **The notifier structurally never posts a verdict.** The PR's own "post-merge: first comment appears" criterion is therefore predicted to fail.

**Severity rationale (Sabine sanity):** this is not a rare edge case — it is the *normal* path on every invocation. The feature's single job (post a verdict) cannot execute.

**Fix options (any one):**
- **(a) Filter out the notifier's own check** before counting, by name:
  ```bash
  SELF="notify"   # or the full check-run name GitHub assigns
  RUNS=$(gh api ".../check-runs?per_page=100" --paginate \
    --jq "[.check_runs[] | select(.name != \"$SELF\") | {name,status,conclusion,html_url}]")
  ```
  Caveat: the check-run name GitHub assigns to an Actions job is the **job name** (here `notify`) — but if a check-run name collides with a real check, this over-filters. Recommend a distinctive job name (e.g. `pr-check-status-notify`) and filter on exactly that. **Verify the exact assigned name post-merge.**
- **(b)** Switch trigger to `check_run: completed` and exclude `github.event.check_run.name == <self>` AND still filter self from the per-SHA query (the query problem persists regardless of trigger).
- **(c)** Query check **suites** for the SHA and a separate combined-status, excluding the Actions suite that is the notifier — more complex; (a) is simplest.

Whichever fix, add an explicit **self-exclusion test** to the post-merge plan: confirm a verdict comment actually appears (current plan asserts this but the design prevents it).

---

### 🟢 Finding #2 — Recursion is bounded (focus #2 reasoning confirmed, with a correction)

The proposed reasoning ("the loop terminates after one extra firing because PATCH doesn't create a new check run") is **directionally right that there is no infinite loop, but the actual stopping mechanism is different** — worth stating precisely:

- The upsert PATCH/POST is idempotent and does not itself create check runs, so the *comment* side cannot spiral. Correct.
- But the notifier **run** completing *does* complete a check_suite, which *would* fire `check_suite:completed` again. The thing that stops re-triggering is **GitHub's built-in recursion guard**: events generated using the default `GITHUB_TOKEN` (and Actions-created check_suite/check_run completions) **do not trigger new workflow runs**. So the notifier's own completion will not spawn another notifier. **Bounded — no loop.** ✅
- Implication: the notifier is only ever triggered by *real* (non-token) check suites. Good — but this does **not** rescue Finding #1, because the *self* check run is still present in the per-SHA REST list at query time regardless of what triggered the run.

---

### 🟡 Finding #3 — Fork PRs (focus #1): `pull_requests[0]` empty for forks

The PR body's claim is accurate: `check_suite` events run the workflow from the **default branch**. Separately, the well-known GitHub quirk holds: `check_suite.pull_requests` is **empty for PRs from forks** (it only populates for same-repo branches). The `if:` guard `pull_requests[0] != null` therefore silently skips all fork PRs.

**For this repo** (solo maintainer + agents, same-repo branches only per CLAUDE.md) this is **acceptable**, but it should be a documented limitation in the header comment so a future external contributor's PR silently getting no verdict isn't a mystery. **NOTE, not blocking.**

---

### 🟢 Finding #4 — Injection (focus #4): safe

- Check names reach markdown via `jq -r '... \(.name) ...'` into `$LIST`, then into `printf` as a positional `%s` arg. No `eval`, no shell re-parse of names → **command injection impossible.** ✅
- Markdown injection: a hostile check name could contain markdown (e.g. `](evil)`), but check-run names are author-controlled only by someone who can add a workflow — same trust level as the maintainer here. **Benign for this threat model.** ✅
- `html_url` comes from the API, not user free-text. ✅

---

### 🟡 Finding #5 — per-SHA vs per-suite scoping (the root cause behind #1)

The design question Gemini is asked to weigh in on: `commits/{sha}/check-runs` aggregates **all** suites for the SHA. That is *desirable* for the verdict (you want all checks, not one suite) — but it is also *why* the notifier sees itself. The right shape is "all check runs for the SHA **minus the notifier's own**." Keep the per-SHA query; add the self-filter. (Documented in #1 fix (a).)

---

### 🟢 Finding #6 — printf body fix (focus #6): confirmed clean

Commit f43265b assembles both bodies with `printf '...\n...'` and `%s` substitution rather than YAML multi-line literals. No run-block indentation leaks into the body, so no spurious ```` ``` ```` code-block rendering. The `## ✅`/`## ❌` headings and `- ❌` list items are column-0 in the produced string. ✅ One micro-nit: the literal `"CI green"` double-quotes inside the green-path `printf` format string are fine (single-quoted format string), but ensure no shell sees them — they don't, they're inside `'...'`. ✅

---

### 🟢 Finding #7 — Permissions: minimal & correct

`pull-requests: write` (comment upsert), `checks: read` (read check runs), `contents: read` (default-token baseline). No secrets. Could arguably drop `contents: read` but it is harmless and often needed implicitly. ✅

---

### 🟡 Finding #8 — Pagination / upsert correctness (minor)

- `--paginate` on both `check-runs` and `comments` — correct for >100 items. ✅
- Upsert marker match: `select(.body | startswith($MARKER))` + `(first // {}) | .id // empty` — robust; picks the first marked comment. If two marked comments ever exist (e.g. a prior race), only the first is updated and the second goes stale. Low risk given single-writer. **NOTE.**
- **Race caveat:** two real suites completing within seconds of each other each fire `check_suite:completed` → two concurrent notifier runs may both pass the (post-fix) pending check and both POST before either PATCHes → duplicate comments on first creation. `concurrency:` group keyed on the PR would harden this:
  ```yaml
  concurrency:
    group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number }}
    cancel-in-progress: false
  ```
  Recommend adding alongside the #1 fix. **NOTE.**

---

### Summary

The feature is well-conceived (canonical machine-stamped verdict, edit-in-place, evidence-first), permissions are minimal, injection-safe, and the printf fix is clean. **But as written it will never post**, because its own Actions check run counts as a pending check under the head SHA (Finding #1). Recursion is safely bounded by GitHub's token guard. Fix #1 (self-filter by job/check name) + add `concurrency:` + document the fork limitation, then re-review.

**BLOCKING:** Finding #1.

— Knuth (lead) / Sabine (sanity), QBP Red Team

JamesPagetButler:
## Gemini Review (REAL API, gemini-3-pro-preview) — PR #508 (PR check-status notifier)

**Personas:** Furey + Feynman | **Tier:** 2 | **Session:** review-20260604-222607
**Scope:** `.github/workflows/pr-check-status.yml` | Independent read, sequential after Red Team.

### Verdict: 🔴 REQUEST CHANGES — structurally inoperable as written

---

### Accuracy — 🔴 BLOCKING: Self-Counting / Never-Posts Bug
The GitHub Actions runner registers this workflow's `notify` job as a check run attached to `$SHA` the moment it queues. Because the job is actively running while it queries `repos/$REPO/commits/$SHA/check-runs`, it appears in the payload with `status: in_progress`. Consequently `PENDING` is **always ≥ 1**. The guard `if [ "$PENDING" -gt 0 ]; then exit 0; fi` evaluates true on every single run — it silently exits and **never posts the comment.** (Independently confirms Red Team Finding #1.)

### Trigger & Recursion Semantics
- **Recursion:** Safe. Actions triggered by the default `GITHUB_TOKEN` do not emit downstream `check_suite` events → no infinite loops. (Confirms Red Team #2.)
- **Forks:** `github.event.check_suite.pull_requests` is always empty for fork PRs. Same-repo workflow → operates correctly here, but this limitation will silently drop fork checks if the external-contribution model ever changes. (Confirms Red Team #3.)

### Completeness — Concurrency Race
If two check suites complete simultaneously, two workflow runs trigger. Once the self-counting bug is fixed, both could evaluate `PENDING == 0` and both find `$CID` empty → both POST → **duplicate bot comments.** (Confirms Red Team #8.)

### Injection
Low but present. `jq -r` passing `(.name)` into a Markdown string means a strangely-named check run (`]`, HTML) could break link formatting. Check names are repo-controlled → acceptable risk, but strict sanitation is skipped. (Agrees with Red Team #4: command injection impossible; markdown injection benign for this trust model.)

### Suggestions
1. **Fix the self-counting bug (BLOCKING)** — exclude this workflow's own check run by name:
   ```bash
   RUNS=$(gh api "repos/$REPO/commits/$SHA/check-runs?per_page=100" --paginate \
          --jq '[.check_runs[] | select(.name != "notify" and .name != "PR check-status notifier") | {name, status, conclusion, html_url}]')
   ```
   (Verify the exact assigned check name post-merge — GitHub uses the job name or workflow name.)
2. **Add concurrency control (strongly recommended):**
   ```yaml
   jobs:
     notify:
       concurrency:
         group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number }}
         cancel-in-progress: false  # queue, don't cancel
   ```
3. **Design scope:** Querying per-SHA (`commits/$SHA/check-runs`) is **definitively the correct approach** over scoping to the specific `check_suite` — PR checks are aggregate state; suite-scoping misses holistic PR status. Keep the per-SHA query, just add the exclusion filter. (Answers the design question: per-SHA-minus-self is right.)

### Overall
**REQUEST CHANGES.** Well-designed (the `printf`-vs-multiline-indentation comments are excellent), but structurally inoperable due to the blocking self-counting defect. Implement the `jq` self-exclusion filter and add a concurrency block.

— Furey / Feynman, via Gemini 3 Pro (real API)

JamesPagetButler:
## Round-1 fixes applied — 627bdf5 + 13c3b58

| Finding | Fix |
|---|---|
| 🔴 #1 self-counting (never posts) — found independently by BOTH reviewers | explicit job name `pr-check-status-notify` + name-based self-exclusion in the per-SHA query; jq self-filter unit-tested locally (mixed pending-self fixture → 0 pending) |
| 🟡 #8 / pagination | `--paginate --jq` emits one array per page → slurp+`add` before counting (both queries) |
| 🟡 concurrency race (both reviewers) | `concurrency:` group keyed per PR number, `cancel-in-progress: true` — also kills the stale-SHA-overwrites-newer-verdict cross-push race |
| 🟡 #3 fork PRs | documented in header as a known limitation (acceptable: same-repo branches only) |
| 🟢 #2 recursion | bounded by the GITHUB_TOKEN guard per review correction — comment updated accordingly in spirit (no code change needed) |

**Post-merge verification added to the plan (per Red Team):** confirm the GitHub-assigned check-run name equals `pr-check-status-notify` on the first live PR, and confirm a verdict comment actually appears + upserts (the self-exclusion makes this possible; round-0 design structurally could not post).

Ready for the round-2 delta review.

— qbp-oppenheimer session

JamesPagetButler:
## 🔴 Red Team — Round-2 delta review (Knuth lead)

**Scope:** delta over round-1 fixes `627bdf5` + `13c3b58`. Verifying the four mechanical fixes against the *current* `.github/workflows/pr-check-status.yml` (re-read on the PR branch, HEAD `13c3b58`). Behavioral facts below were checked against `jq` directly, not asserted from memory.

### Verdict: ✅ APPROVE

The round-1 blocker (self-counting deadlock) is correctly closed and the three supporting fixes are sound. One non-blocking hardening note on query 1.

---

### Round-1 blocker resolution

| # | Round-1 finding | Fix | Verified |
|---|---|---|---|
| 1 | Self-counting: notifier's own `in_progress` check run keeps PENDING > 0 forever → never posts | name-based self-exclusion | ✅ PASS |
| 2 | `--paginate --jq` emits one array per page; counting the concatenated docs is wrong | slurp `-s 'add'` flatten | ✅ PASS (with note) |
| 3 | Per-SHA concurrency lets a stale-SHA run overwrite a newer verdict | PR-keyed concurrency + cancel-in-progress | ✅ PASS |
| 4 | Fork PRs silently get no comment | documented as KNOWN LIMITATION | ✅ PASS |

---

### Finding-by-finding

**1. Self-filter name match (the typo-class risk).** The job declares its name on line 49:

```yaml
    name: pr-check-status-notify
```

and the filter binds the identical literal on line 70 and applies it on line 73:

```bash
          SELF="pr-check-status-notify"
          ... | jq -s 'add | map(select(.name != "'"$SELF"'"))')
```

Byte-for-byte identical. A mismatch here would silently resurrect the deadlock, so this is the load-bearing line of the whole fix — it is correct. ✅

**2. Timeline walk with the filter in place.** On `check_suite.completed`, the job queries all check-runs for `$SHA`, drops every run named `pr-check-status-notify`, then counts non-`completed`. The filter is **name-based, not id-based**, so it excludes *all* notifier instances for that SHA — including a sibling notifier instance spawned by a different suite's completion on the same SHA. Combined with `cancel-in-progress: true` (only one notifier survives per PR anyway), there is no path where a notifier counts itself or a peer notifier as pending. The post fires iff every non-notifier run for the SHA is `completed`. ✅

**3. Concurrency expression validity.** Line 40:

```yaml
  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.check_suite.head_sha }}
```

`||` is valid inside `${{ }}` and is value-coalescing (returns the first operand that is not null/false/0/''), not strictly boolean — so `number || head_sha` yields `number` when present, else `head_sha`. This parses and behaves as intended; it does **not** error at workflow load. ✅

**4. Slurp on a possibly-empty stream — the one real subtlety.** I reproduced the failure mode directly:

```
$ printf '' | jq -s 'add | map(select(.name!="x"))'
jq: error: Cannot iterate over null (null)   # exit 5
```

So *if* the slurped stream were ever empty, `add` → `null` and `map` over `null` aborts the job under `set -euo pipefail`. **However**, this cannot fire here, because:

- Both per-page `--jq` filters are wrapped in `[...]` (`[.check_runs[] | …]` and `[.[] | select(…)]`), so each page emits exactly one array (`[]` at worst), never zero values.
- `gh api --paginate` always issues ≥1 request → ≥1 document → the slurp always sees ≥1 array → `add` over `[[]]` = `[]`. Verified:

```
$ printf '[]' | jq -s 'add | map(select(.name!="x"))'   →  []
```

Query 2 (`add | (first // {}) | .id // empty`) is additionally self-defended by `(first // {})`, so even a hypothetical empty slurp degrades to `empty`, not an error — verified safe.

**Conclusion: not a live defect.** But the query-1 safety rests entirely on nobody dropping the `[...]` wrapper in a future edit; the moment that wrapper goes, query 1 silently reverts to the `null`-iteration crash. Cheap insurance, **non-blocking**: change line 73 to `jq -s 'add // [] | map(select(...))'`. This makes correctness independent of the upstream filter shape and matches the defensive `// {}` already used on query 2. Recommend, do not block.

**5. cancel-in-progress duplicate-comment window (minor, noted).** A cancellation that lands *between* the upsert's read (`CID` lookup) and its write (POST when `CID` empty) could in principle let two instances both decide "no existing comment" and both POST → duplicate. In practice `cancel-in-progress` cancels *older* runs at the point a newer one starts, and the upsert read→write window is sub-second, so the race is vanishingly small and self-heals on the next completion (the next run finds the first comment and PATCHes it, leaving at most one transient dupe). Acceptable for a solo-maintainer repo. Noted, **not blocking**.

---

### Acceptance criteria
This is an infra/tooling PR with no linked issue carrying formal `- [ ] **AC:**` checkboxes (header-documented behavior contract instead). The behavior contract in the file header (lines 11–17) is satisfied by the current implementation. No AC gate to fail.

**Knuth:** the proof obligation — "PENDING reaches 0" — now holds because the notifier is provably excluded from its own census. The empty-stream lemma is the only loose thread, and it is currently discharged by the `[...]` invariant; I'd pin it explicitly with `add // []` so the proof doesn't depend on a filter written elsewhere in the same line. APPROVE.

JamesPagetButler:
## 🔵 Gemini (Furey / Feynman) — Round-2 delta review · REAL API

**Model:** `gemini-3-pro-preview` · **Session:** `review-20260604-223145`
**Session note:** the round-1 session (`review-20260604-222607`) could **not** be resumed (CLI session-ID collision, exit 42); this delta was run on a **fresh session** with the full round-1 context re-stated in the prompt. Continuity is by restated context, not CLI session state.

### Single verdict: ✅ APPROVE

| # | Review item | Status | Finding |
|---|---|---|---|
| 1 | Self-filter name match | ✅ PASS | `name: pr-check-status-notify` exactly matches `SELF="pr-check-status-notify"`. No typos. Deadlock resolved. |
| 2 | Timeline walk | ✅ PASS | `select(.name != …)` drops **all** sibling notifier instances across all suites — it keys on the run *name*, not a run ID. Post fires iff all non-notifier runs are completed. |
| 3 | Concurrency expression | ✅ PASS | `${{ … \|\| … }}` coalescing is valid GitHub Actions syntax; safely falls back to `head_sha` when `pull_requests[0]` is null. Does not error at parse. |
| 4 | Slurp empty-stream edge | ⚠️ NOTE | Safe (not a live defect). Per-page queries wrapped in `[...]` → `gh api` always emits ≥ `[]` → `jq -s 'add'` sees `[[]]` → `[]`, so `map` never hits the null-iteration error. Latent fragility only. |
| 5 | cancel-in-progress window | ⚠️ NOTE | A narrow race exists between CID lookup and POST/PATCH if a cancel lands exactly there → possible duplicate comment. Acceptable edge case. |

---

**Accuracy.** Resolves the round-1 feedback perfectly. The self-exclusion targets the exact job-name string; `||` in `${{ }}` is a valid, robust concurrency-group fallback; the `jq` pipeline counts pending runs and parses failures without crashing on empty datasets, thanks to the per-page `[...]` array wrappers.

**Completeness.** Fulfills the behavior contract in the file header (lines 11–17). The documented fork-PR limitation (lines 20–23) appropriately brackets expectations. No formal AC checkboxes exist; infra/tooling standards are met.

**Suggestion (non-blocking hardening).** Protect against a future edit accidentally dropping the `[...]` wrappers — make correctness independent of the upstream filter shape:
- Line 73: `| jq -s '(add // []) | map(select(.name != "'"$SELF"'"))')`
- Line 104: `| jq -s '(add // []) | (first // {}) | .id // empty' -r)`

**Overall.** APPROVE. The changes resolve the round-1 self-counting deadlock and safely handle concurrency and edge-case payload parsing. Production-ready.

— Furey / Feynman, via real Gemini API (`gemini-3-pro-preview`)

Issue #507, opened 2026-06-04
[process] Deep enforcement for evidence-first rules: title linting + oracle fixture premise-audit tooling


JamesPagetButler:
Follow-up from PR #505 Gemini review (blockers G-4/G-5 adopted minimally as reviewer-side rules; this tracks the mechanical versions).

1. **Title/commit-message truth-in-labelling lint:** CI check flagging verdict-words ('proves', 'establishes', 'confirms') in PR titles for theory-bearing diffs unless the body's headline section carries a verdict — crude heuristic, reviewer remains authority (rule 6, review_tiers.md).
2. **Oracle fixture premise audit:** mechanical check that any 'oracle run' evidence link in a PR body references a fixture/config file committed in-repo at a pinned SHA — no out-of-band JSON. Pairs with the differential guard (#504).

**AC:**
- [ ] Both checks exist, are advisory-or-blocking per beekeeper choice
- [ ] Documented in review_tiers.md evidence-first section

## 3. workflow_run enumeration-drift lint (from PR #511 round-2 reviews, non-blocking)
CI check asserting `.github/workflows/pr-check-status.yml` `workflow_run.workflows` stays in sync with `grep ^name: .github/workflows/*.yml` (minus the notifier itself) — a new workflow not appended means its completion cannot finalize a verdict.

## 4. Box-state gate on the ALL-GREEN verdict (from qbp-architecture seq=553 + the #513/#234 unticked-box incidents, 2026-06-05)
ALL-GREEN additionally requires zero unticked merge-blocking boxes (parsing github-best-practices §2.2 format: "Sign-off required (merge-blocking)" vs "Informational" sections). New intermediate state: CI-green + unticked sign-off boxes → nudge the author/reader (citing any posted APPROVE comment as the evidence — auto-nudge, not auto-tick), do NOT escalate merge-ready to beekeeper. Contract format owned by qbp-architecture; QBP notifier implementation owned by qbp-oppenheimer; federation adopts by copy.

**AC:**
- [ ] ALL-GREEN gated on box-state per §2.2 sections
- [ ] nudge state implemented (comment-evidence-vs-unticked-box detection)
- [ ] adopted format documented in the workflow header

## 5. Box semantics: single-select groups + self-ticking evidence boxes (from the #518 routing-axis confusion, 2026-06-06)
[...]

PR #510, opened 2026-06-04
[validation] #492: single-source expectation emission on both backends + pre-commit differential guard

JamesPagetButler:
## Summary

Closes the loop the differential bridge opened on its first flight: **the same dual-float-path bug existed in both backends**, and this PR fixes both with one rule + deploys the guard that prevents recurrence.

| Piece | Change |
|---|---|
| QBP oracle (`proofs/QBP/Oracle/Main.lean`) | `expectation := round6(prob_up) − round6(prob_down)` — single source of truth; the independent cos-θ float path eliminated (6 fixture rows were internally inconsistent, not 4 as #492 estimated) |
| Bridge oracle (`PhysleanBridge/Oracle.lean`) | same rule, same convention comment (beekeeper path-1 ruling): *expectation is a DERIVED field; the physics comparison lives in the independently-derived probabilities*. The PROVEN cos-θ law (`expectation_eq`) untouched |
| `tests/oracle_predictions.json` | regenerated; byte-idempotent under `lake exe oracle` |
| `scripts/check_physlean_differential.py` + `.pre-commit-config.yaml` | the differential guard: prebuilt bridge exe vs committed fixture (~2 s), scoped to oracle-relevant paths; loud WARN + pass when bridge unbuilt (never blocks a commit on an 8 GB build); `--self-test` corruption mode |

## Mechanical verification evidence

- [x] **GREEN:** differential PASS exit 0 — **10/10 independent physical derivations + 5/5 expectation consistency checks** (per the #492 convention, expectation is definitionally shared; independence lives in prob_up/prob_down — see the rescoped INDEPENDENCE CLAIM in diff_oracle.py) — 45° row healed (`0.707106 == 0.707106`, absdiff 0.00e+00); was the residual 1-ULP RED
- [x] **RED:** `--self-test` (+0.001 corruption at 45°/prob_up) → FAIL, exit 1 — the guard's green means something
- [x] **ABSENT:** missing bridge → loud WARN, exit 0, no build triggered
- [x] `lake build QBP` green; bridge `lake build physlean_oracle` green (1:34 incremental — hardlink-cloned cache, no 8 GB rebuild); 86/86 differential tests, 383/383 full pytest
- [x] `#print axioms` all 6 bridge decls + QBP audited theorems: ⊆ {propext, Classical.choice, Quot.sound}
- [x] Zero sorry / native_decide / vacuous-True in touched Lean; bridge pin unchanged (v4.29.1)
[...]

PR #511, opened 2026-06-04
[infra] Two-phase review trigger: tier-label on open → tech-green review-ready → patch-id freshness at all-green

JamesPagetButler:
## Summary

Closes the **initiation half** of the review loop (the gates from #495 close the completion half). Extends the #508 notifier into the beekeeper-ratified two-phase state machine:

```
PR opened/synced ─► tier label auto-applied from paths        (Phase 0, deterministic)
checks run ─► TECH-GREEN (excl. review gates) ─► review-ready label + checklist
reviews posted ─► gates green ─► ALL-GREEN + patch-id freshness ─► merge-ready verdict
fix breaks CI ─► RED, review-ready REVOKED, cycle re-arms
```

| Piece | Mechanism |
|---|---|
| Tier label on open | classifies from changed paths (mirrors `check_tier3_label.py` globs); never overrides a present tier label |
| Tech-green threshold | all checks green **excluding** review-completeness gates — naive "all green" deadlocks because the label gate can't pass before reviews exist |
| Review checklist | posted in the verdict comment with the `<!-- reviewed-sha: … -->` marker convention (Red Team → Gemini → synthesis → HVR) |
| Regression revocation | any tech check fails after a fix push → `review-ready` removed, "not reviewable until tech-green again" |
| **Review freshness** | at all-green, compares `git patch-id --stable` of last-reviewed SHA vs head: identical (rebase/empty retrigger) = fresh; content change = **🔶 DELTA ROUND REQUIRED — do not merge on these reviews**; no marker = warned unverifiable (grandfathers pre-convention PRs) |

**What stays human/orchestrated:** review *judgment* (dispatch with tailored probes) and the merge itself. This automates initiation and staleness detection only — the PATTERN-01 register is entirely forgotten-initiation incidents; this closes that class mechanically.

- [x] YAML valid (`yaml.safe_load`); embedded bash syntax-checked (`bash -n`)
- [x] Self-counting fix preserved (job-name self-filter); pagination slurp hardened with `add // []` (the #508 round-2 non-blocking suggestion, folded in)
- [ ] CI green on this PR
- [ ] Post-merge live test: next PR gets tier label + phase verdicts; a deliberate fix-push exercises revocation + delta-round detection

## CTH anchor impact
- [x] N/A — CI workflow only.

Refs #508 (extends), #495 (completes the loop it started), #507 (deep enforcement), PATTERN-01 (initiation class).

### DIFF
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -63,15 +63,23 @@ permissions:
   checks: read
   contents: read
 
-concurrency:
-  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.workflow_run.pull_requests[0].number || github.event.pull_request.number || github.event.check_suite.head_sha || github.event.workflow_run.head_sha }}
-  cancel-in-progress: true
+# CONCURRENCY (review round-1, Gemini blocking finding): the two jobs are
+# separate AXES and must never share a cancellation group — a slow OLD-SHA
+# suite completion firing notify could cancel a NEW-SHA tier-label run (the
+# new commit would silently miss its tier classification). Job-level groups
+# (below) separate the axes; within each axis, stale runs still cancel.
+# (Job-level rather than event_name-keyed: after the workflow_run trigger fix,
+# notify fires from TWO event types — event_name keying would let two notify
+# instances race the comment upsert.)
 
 jobs:
   # ── Phase 0: tier label from changed paths (instant, deterministic) ─────────
   tier-label:
     name: pr-tier-label
     if: ${{ github.event_name == 'pull_request' }}
+    concurrency:
+      group: pr-tier-label-${{ github.event.pull_request.number }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       - name: Classify tier from changed paths and apply label if absent
@@ -106,6 +114,9 @@ jobs:
     # below excludes it by name (self-counting fix, #508 review finding #1).
     name: pr-check-status-notify
     if: ${{ (github.event_name == 'check_suite' && github.event.check_suite.pull_requests[0] != null) || (github.event_name == 'workflow_run' && github.event.workflow_run.pull_requests[0] != null) }}
+    concurrency:
+      group: pr-notify-${{ github.event.check_suite.pull_requests[0].number || github.event.workflow_run.pull_requests[0].number }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       # full history so patch-ids can be computed for both head and the
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -29,7 +29,29 @@
 
 name: PR check-status notifier
 
+# TRIGGER DESIGN (post-#508 live-test finding): check_suite:completed events
+# from ACTIONS-CREATED suites never trigger workflows (GitHub's recursion
+# guard) — on its first live PR the #508 notifier only fired on CodeQL's
+# suites and structurally could not see the last Actions check finish.
+# workflow_run:completed is the supported channel for Actions workflows; the
+# check_suite trigger is kept for non-Actions apps (CodeQL etc.).
+# MAINTENANCE: a NEW workflow added to the repo must be appended to the
+# workflow_run list below, or its completion won't re-evaluate the verdict
+# (a later-finishing unlisted workflow = missed final event).
 on:
+  workflow_run:
+    workflows:
+      - Lean Foundations Gate
+      - CTH Schema Lint
+      - Link Checker
+      - CTH Anchor Impact
+      - CI
+      - Tier-3 Review Gate
+      - Differential Test (Phase 4d)
+      - Lint
+      - Tenancy XRef Lint
+      - Validate JSON
+    types: [completed]
   check_suite:
     types: [completed]
   pull_request:
@@ -42,7 +64,7 @@ permissions:
   contents: read
 
 concurrency:
-  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.pull_request.number || github.event.check_suite.head_sha }}
+  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.workflow_run.pull_requests[0].number || github.event.pull_request.number || github.event.check_suite.head_sha || github.event.workflow_run.head_sha }}
   cancel-in-progress: true
 
 jobs:
@@ -83,7 +105,7 @@ jobs:
     # IMPORTANT: this job name is the notifier's own check-run name; the query
     # below excludes it by name (self-counting fix, #508 review finding #1).
     name: pr-check-status-notify
-    if: ${{ github.event_name == 'check_suite' && github.event.check_suite.pull_requests[0] != null }}
+    if: ${{ (github.event_name == 'check_suite' && github.event.check_suite.pull_requests[0] != null) || (github.event_name == 'workflow_run' && github.event.workflow_run.pull_requests[0] != null) }}
     runs-on: ubuntu-latest
     steps:
       # full history so patch-ids can be computed for both head and the
@@ -95,8 +117,8 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
           REPO: ${{ github.repository }}
-          PR: ${{ github.event.check_suite.pull_requests[0].number }}
-          SHA: ${{ github.event.check_suite.head_sha }}
+          PR: ${{ github.event.check_suite.pull_requests[0].number || github.event.workflow_run.pull_requests[0].number }}
+          SHA: ${{ github.event.check_suite.head_sha || github.event.workflow_run.head_sha }}
         run: |
           set -euo pipefail
           MARKER="<!-- pr-check-status-notifier -->"
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -1,57 +1,97 @@
-# PR check-status notifier — posts ONE upserted comment per PR with the final
-# check verdict for the current head SHA, timestamped.
+# PR check-status notifier + two-phase review trigger.
 #
-# Why (2026-06-04, two incidents in one day): humans and agents kept acting on
-# STALE check states — a "ready to merge" claim made without re-querying, and a
-# red verdict read after the gate had already re-run green. This makes the PR
-# conversation itself carry a machine-stamped verdict for the current head, so
-# "CI green" claims have a canonical artifact to link (evidence-first rule,
-# docs/workflows/review_tiers.md).
+# Phase model (beekeeper-ratified design, 2026-06-04):
 #
-# Behavior:
-#   - fires when any check suite completes
-#   - if other suites for the same SHA are still running -> exits quietly
-#   - else upserts (edit-in-place, never spam) a single marked comment:
-#       ALL GREEN (N checks) @ timestamp for sha
-#       or FAILED: <names + links> @ timestamp for sha
-#   - new push -> next completion updates the comment for the new SHA
+#   PR opened/synced ──► tier label auto-applied from changed paths (job: tier-label)
+#   checks run       ──► TECH-GREEN: all checks green EXCLUDING review-completeness
+#                        gates ──► `review-ready` label + review checklist posted
+#                        (reviews are dispatched by the orchestrator; initiation is
+#                        now mechanical, not memory)
+#   reviews posted   ──► review gates turn green ──► ALL-GREEN + review-freshness
+#                        check (patch-id) ──► MERGE-READY verdict (merge stays the
+#                        beekeeper's call)
+#   fix breaks CI    ──► RED verdict, `review-ready` REVOKED, cycle re-arms
+#
+# Review freshness: review/fix comments include a machine-readable marker
+#   <!-- reviewed-sha: <full-sha> -->
+# The ALL-GREEN verdict compares git patch-id of the last reviewed SHA vs head:
+# identical patch-id (rebases / empty retrigger commits) keeps reviews fresh;
+# a content change demands a delta round. No marker -> freshness UNVERIFIABLE
+# (warned, not blocked — grandfathers pre-convention PRs).
+#
+# Why two thresholds: "trigger reviews when ALL checks pass" deadlocks — the
+# tier-3 review-label gate is itself a check that stays red until reviews
+# exist. TECH-GREEN excludes review gates; ALL-GREEN includes them.
+#
+# KNOWN LIMITATION: check_suite payloads carry an empty pull_requests[] for
+# FORK PRs, so fork PRs get no verdict comment (acceptable: solo maintainer +
+# agent sessions, same-repo branches only).
 
 name: PR check-status notifier
-
-# KNOWN LIMITATION: check_suite payloads carry an empty pull_requests[] for
-# FORK PRs, so fork PRs get no verdict comment (acceptable for this repo:
-# solo maintainer + agent sessions, same-repo branches only — documented per
-# review finding #3 so a future external PR's silence isn't a mystery).
 
 on:
   check_suite:
     types: [completed]
+  pull_request:
+    types: [opened, reopened, synchronize]
 
 permissions:
   pull-requests: write
+  issues: write
   checks: read
   contents: read
 
-# collapse simultaneous firings per PR (not per SHA: PR-keying also cancels a
-# stale in-flight run for an OLD head SHA when a new push's suites complete,
-# so an outdated verdict can never overwrite a newer one — review concurrency
-# finding, round 1)
 concurrency:
-  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.check_suite.head_sha }}
+  group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number || github.event.pull_request.number || github.event.check_suite.head_sha }}
   cancel-in-progress: true
 
 jobs:
-  notify:
-    # IMPORTANT: this job name is the notifier's own check-run name; the query
-    # below excludes it by name so the notifier never counts itself as pending
-    # (review finding #1 — without the self-filter, PENDING never reaches 0
-    # and the notifier structurally never posts).
-    name: pr-check-status-notify
-    # only suites attached to a PR
-    if: ${{ github.event.check_suite.pull_requests[0] != null }}
+  # ── Phase 0: tier label from changed paths (instant, deterministic) ─────────
+  tier-label:
+    name: pr-tier-label
+    if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Compose and upsert verdict comment
+      - name: Classify tier from changed paths and apply label if absent
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          LABELS=$(gh pr view "$PR" --repo "$REPO" --json labels --jq '[.labels[].name]')
+          # never override a human/orchestrator tier choice
+          if echo "$LABELS" | jq -e 'map(select(test("^tier-[123]-review$"))) | length > 0' > /dev/null; then
+            echo "tier label already present — leaving as-is"; exit 0
+          fi
+          FILES=$(gh pr view "$PR" --repo "$REPO" --json files --jq '.files[].path')
+          # theory-bearing globs mirror scripts/check_tier3_label.py THEORY_GLOBS
+          TIER="tier-1-review"
+          while IFS= read -r f; do
+            case "$f" in
+              proofs/*.lean|archive/cth-inventory/*.json|paper/*|docs/foundations/*)
+                TIER="tier-3-review"; break ;;
+              *.py|*.lean|*.yml|*.yaml|*.json|*.sh|*.toml|tests/*|scripts/*|proofs/*|src/*)
+                TIER="tier-2-review" ;;
+            esac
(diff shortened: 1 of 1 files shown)


JamesPagetButler:
## Red Team Review — PR #511 (Knuth lead; Sabine, Grothendieck)

**Tier:** 2 (infra; CI workflow, no theory artifacts). **Scope:** `.github/workflows/pr-check-status.yml` rewritten into the two-phase review-trigger state machine.

Mechanical verification performed locally on the PR branch head `33f90e8`: `yaml.safe_load` parses; both `run:` blocks extracted and `bash -n` clean; tier-glob behavior tested empirically against `scripts/check_tier3_label.py`; `fnmatch` semantics verified in a real interpreter.

### Verdict table

| # | Probe | Verdict | Severity |
|---|---|---|---|
| 1 | Concurrency group / event-context coalescing | PASS (reasoning confirmed) | — |
| 2 | Self-counting regression (tier-label run in PENDING) | PASS — self-healing, verified | — |
| 3 | Patch-id base-drift → false DELTA ROUND | PASS — safe-direction error, ruled acceptable | NIT |
| 4 | `review-ready` label churn idempotency | PASS | — |
| 5 | Marker regex + `last`-in-creation-order | PASS with one edited-comment caveat | NIT |
| 6 | Tier classifier glob divergence vs `check_tier3_label.py` | **CORRECTION: divergence premise is false — they agree** | NIT (redundant glob; broad-by-design ruled OK) |
| 7 | AC / test-plan / YAML+bash validity | PARTIAL — live-test AC unchecked (expected, post-merge) | NON-BLOCKING |

**Overall: APPROVE.** No blocking findings. The two prior structural bugs (self-counting, pagination slurp) both survived the rewrite intact. Detail below; note the substantive correction on probe 6.

---

### 1. Event/concurrency interplay — PASS

Reasoning on the coalescing expression is correct:

```yaml
group: pr-check-status-${{ github.event.check_suite.pull_requests[0].number
       || github.event.pull_request.number
       || github.event.check_suite.head_sha }}
```

- **Null-property access does not error.** In Actions expression syntax, indexing/property-accessing a `null` context yields `null` (not an evaluation error), and `||` coalesces falsy → next operand. For a `pull_request` event, `check_suite` is `null`, `check_suite.pull_requests[0].number` → `null`, falls through to `pull_request.number`. For a `check_suite` event, the first operand resolves and `pull_request.number` is never consulted. Both events therefore key on **the same PR number** — which is the intent.
- **Cross-event cancellation timeline.** Because tier-label and notify share the group with `cancel-in-progress: true`, I walked the synchronize race:
  - `synchronize` (new push) fires both a `pull_request` event (→ tier-label) and, seconds-to-minutes later, suite completions (→ notify) **for the new SHA**.
  - tier-label runs in seconds and exits; an in-flight notify for the **OLD** SHA being cancelled by the new tier-label run is **desirable** (the old verdict is stale).
  - The ordering concern — *could a tier-label run starting AFTER the new-SHA notify has begun cancel that notify?* — does not bite in practice: tier-label is triggered by the `synchronize` event which is emitted **before** the new suites even start, let alone complete. So tier-label is essentially always the *earlier* arrival; by the time notify (suite-completed) starts, tier-label has finished and is not "in progress" to be cancelled by, nor to cancel, the notify. The window where tier-label starts after notify is not reachable on the normal GitHub event timeline.
  - Residual: GitHub does not guarantee event delivery order under retries/backpressure. The blast radius if it ever did mis-order is **one missed verdict comment**, self-healed by the next suite completion (the upsert is idempotent on the marker). Acceptable. **No action.**

### 2. Self-counting regression — PASS (self-healing, verified)

The `tier-label` job creates a check run (`pr-tier-label`) on the head SHA's suite. It is **not** in `SELF` and **not** in `REVIEW_GATES`, so while it is `in_progress` the notify PENDING count is >0 and notify exits early — correct, no false verdict on a partial set.

The self-heal holds: tier-label completing is itself a `check_suite` state change that ultimately rolls into suite completion; the **last** suite-completed event for the SHA fires after every run (including `pr-tier-label`) is `completed`, at which point PENDING=0 and the verdict posts. There is no ordering in which `pr-tier-label` is the *eternally* last pending check — it terminates in seconds, deterministically, with no dependency that could wedge it. The original #508 self-counting blocker (notify counting *itself*, never reaching PENDING=0) is fully preserved by the `SELF` filter and is unaffected by the new job. **No action.**

### 3. Patch-id base-drift — PASS, ruled acceptable (NIT)

```sh
pid() { git diff "$(git merge-base "origin/$BASEREF" "$1")".."$1" | git patch-id --stable | cut -d' ' -f1; }
```

Correct concern: if `master` advances between the review SHA and head, `merge-base(origin/master, SHA)` differs for the two SHAs, so each diff is taken against a *different* base. If the intervening master commits touch lines near the PR's hunks, `patch-id --stable` can differ even when the PR's own authored content is unchanged → **false DELTA ROUND**.

**Ruling:** this is a **safe-direction error**. A false DELTA ROUND tells a human "re-review before merge" — costly at worst, never unsafe. The dangerous direction (stale review wrongly declared fresh) does not occur from base drift, because drift can only *break* patch-id equality, never manufacture it. The honest framing in the PR ("identical = fresh; different = delta round") is sound under this asymmetry. A future refinement (diff each SHA against *its own* merge-base — which this already does — but additionally rebase-normalize, or compare `git range-diff`) would reduce false positives, but is not owed here. **NIT, non-blocking.** Suggest a one-line code comment noting the base-drift false-positive is intentional safe-direction.

### 4. Label churn idempotency — PASS

`gh pr edit --add-label review-ready` on an already-present label is a no-op success; `--remove-label` on an absent label returns nonzero but is wrapped `|| true`. The helper functions also swallow all errors (`2>/dev/null || true`), so neither the TECH-GREEN re-fire (re-adding) nor the RED re-fire (re-removing) churns or fails the step. **No action.**

### 5. Marker regex + `last` ordering — PASS (NIT)

`capture("<!-- reviewed-sha: (?<s>[0-9a-f]{7,40}) -->")` over `[.[] | .body]` from the comments API (paginated, slurped, flattened) then `| last`. The comments API returns issue comments in **ascending creation order**, so `last` = most-recently-created marker = the latest review/fix stamp. Correct for the intended flow (a fix-comment re-stamps after reviews).

Two caveats, both NIT:
- **Edited comments** retain their original creation-order position but can carry a newer SHA in the edited body. `last` keys on creation order, not edit time — so if a reviewer *edits* an older comment to bump its `reviewed-sha` rather than posting a new one, `last` could miss it. The convention this PR establishes (one marker per fresh post) makes this an off-convention edge; worth a half-sentence in `review_tiers.md` when the convention is documented. Not a code change.
- The regex `[0-9a-f]{7,40}` accepts abbreviated SHAs (≥7). Comparison `[ "$LASTREV" = "$SHA" ]` against the full 40-char head will then fall through to the patch-id branch (since abbreviated ≠ full), which fetches `$LASTREV` and resolves it via git — git accepts the abbrev, so it still works. Fine; just noting the exact-match fast-path is bypassed for abbreviated stamps. **No action.**

### 6. Tier classifier glob divergence — CORRECTION (premise false)

The probe asserted the bash `case` is **broader** than `check_tier3_label.py` because "bash `case *` crosses `/` unlike fnmatch." **This is incorrect, and I verified it in a real interpreter:**

```python
fnmatch.translate('proofs/*.lean')  ->  (?s:proofs/.*\.lean)\Z
fnmatch.fnmatch('proofs/QBP/Foundations/X.lean', 'proofs/*.lean')  ->  True
```

Python's `fnmatch` translates `*` to `.*`, which **does** cross `/` (fnmatch treats `/` as an ordinary character — unlike `glob.glob` or `pathlib`). So bash `case proofs/*.lean` and Python `fnmatch('proofs/*.lean')` **agree**: both match arbitrarily-nested `.lean` under `proofs/`. There is **no broadening divergence** on slash-crossing.

What this *does* reveal: the Python list's `proofs/**/*.lean`, `paper/**`, `docs/foundations/**` entries are **redundant** — each is already subsumed by its `*` sibling under fnmatch semantics. Harmless, but the "mirrors `check_tier3_label.py` THEORY_GLOBS" comment in the workflow is accurate, and the two classifiers produce the **same tier-3 set**. 

One genuine, smaller divergence remains and should be named: the bash classifier's **second** case arm (`*.py|*.lean|*.yml|...|proofs/*|src/*` → tier-2) is a workflow-only addition with no Python counterpart — that is by design (the Python script only decides tier-3 gating; the workflow additionally assigns tier-1/2 as a default). `*.lean` files **outside** the tier-3 paths can't occur (any `.lean` matches `proofs/*.lean`→tier-3 first via `break`), so no `.lean` ever lands in tier-2. Consistent. **NIT only: drop the redundant `**` globs in the Python list during the next housekeeping pass for clarity; flag the corrected fnmatch fact in the review record so it isn't re-litigated.**

### 7. AC / test-plan / validity — PARTIAL (non-blocking)

- YAML valid (`safe_load` OK), both bash blocks `bash -n` clean — PR's mechanical-evidence claims reproduced and confirmed.
- Test-plan checkboxes: two unchecked — `CI green on this PR` and `Post-merge live test`. Both are inherently post-merge/post-CI; their unchecked state is expected, not a gate violation at review time. The live-test item (next PR exercises tier-label + revocation + delta-round) is the **real** validation and should be tracked to closure after merge — recommend a follow-up issue so PATTERN-01 doesn't quietly re-open if live behavior diverges from the dry analysis.
- Self-counting + pagination `add // []` hardening: both present and correct.

---

PR #511 references #508 (extends), #495 (completes), #507, PATTERN-01. No single linked issue carries a discrete `- [ ] **AC:**` checklist that this PR `Closes`; it is an extension PR. The PR's own mechanical-evidence list is the operative AC surface:

| Criterion | Status | Evidence |
|---|---|---|
| YAML valid; bash syntax-checked | PASS | reproduced locally (`safe_load`, `bash -n`) |
| Self-counting fix preserved | PASS | `SELF` filter intact, probe 2 |
| Pagination slurp hardened (`add // []`) | PASS | present at both slurp sites |
| CI green on this PR | DEFERRED | post-CI, not assessable at review time |
| Post-merge live test | DEFERRED | requires merge; track via follow-up issue |

**AC verdict:** PASS on all review-time-assessable criteria; two deferred-by-nature items, neither blocking, one (live test) recommended as a tracked follow-up.

---

**Knuth:** Mechanically sound. The printf-body discipline and single-source `REVIEW_GATES` list are the right call. My only substantive note is the probe-6 correction — the classifiers agree; the redundant `**` globs are cosmetic.
**Sabine:** The patch-id freshness check is the honest part of this design — it fails toward "re-review," never toward "merge stale." That asymmetry is what makes the automation trustworthy. No objection.
**Grothendieck:** Two-threshold (TECH-GREEN ⊂ ALL-GREEN) factoring is the correct resolution of the label-gate deadlock; the structure is clean.

**BLOCKING: none. Verdict: APPROVE.** Recommend a follow-up issue for the post-merge live test and a housekeeping nit to drop redundant `**` globs.

Next in the sequence: Gemini (Furey/Feynman, real API).

JamesPagetButler:
## Gemini Review — PR #511 (REAL API, gemini-3-pro-preview; Furey/Feynman)

Sequential second review (Red Team / Knuth ran first). Independent read. Session: `review-20260604-225506`.

---

## Clarity
The phase model documentation at the top is excellent. The state machine is explicit, and separating the mechanical thresholds (TECH-GREEN vs ALL-GREEN) dissolves the deadlock elegantly.

## Accuracy

**PROBE 1 — Event/concurrency interplay:**
- (a) **Expression Safety:** Safe. GitHub Actions expression syntax handles missing object properties without erroring. It evaluates to null, allowing the `||` coalescing to safely fall back to the pull_request number or head_sha.
- (b) **Cancellation Race:** **REAL RACE.** Because `pull_request` and `check_suite` share the exact same concurrency group (`pr-check-status-<PR_NUMBER>`), a late-arriving `check_suite:completed` event from an *old* commit (e.g., a slow test suite finishing) will cancel the `tier-label` job currently running for a *new* commit triggered by `synchronize`. `tier-label` takes ~5 seconds; if canceled mid-flight, the PR misses its tier classification entirely.

**PROBE 2 — Self-counting regression:**
- **Self-healing.** `notify` ignores itself but correctly counts `pr-tier-label` as a pending check if it's still running. This does *not* wedge the system. Because `tier-label` is itself a GitHub Actions job, its completion will intrinsically emit a final `check_suite:completed` event. That final event acts as a sweep, triggering `notify` one last time where `PENDING` will strictly equal 0. The fix from #508 survived successfully.

**PROBE 3 — Patch-id base-drift:**
- **Safe-direction error only.** `git patch-id --stable` normalizes line numbers but hashes the diff literal. If `master` advances and the PR is cleanly rebased or merged without altering context lines, the `patch-id` remains identical (Fresh). If `master` advances and alters the context lines surrounding the PR's changes, the diff literal changes, triggering a `patch-id` mismatch. This causes a FALSE DELTA ROUND (the authored intent didn't change, but context did). This is a strictly acceptable, safe-direction error that forces a human to bless the context-shifted diff. Base drift cannot produce a dangerous error (falsely stamping a changed PR as fresh) because the diff content would have to collide cryptographically.

**Sanity Checks:**
- Bash `case` `*` matches across directories (like Python's `fnmatch`), so `proofs/*` correctly matches `proofs/lean4/file.lean`. They are in structural agreement.
- The regex extraction and `last` pipe for the marker are chronologically sound since `gh api` paginates chronologically.
- `gh pr edit --add-label` idempotency is correctly handled with `2>/dev/null || true`.

## Completeness
The workflow maps the state machine completely, but conceptually, `tier-label` and `notify` do not belong in the same concurrency group since they operate on entirely different orthogonal axes of the PR lifecycle (setup vs. verdict).

## Suggestions
1. **Split Concurrency Groups (BLOCKING):** Add the event name to the concurrency group to prevent `check_suite` events from canceling `pull_request` setup events:
   `group: pr-check-status-${{ github.event_name }}-${{ github.event.check_suite.pull_requests[0].number || github.event.pull_request.number }}`
   *(Note: You can omit SHA from the fallback. Grouping strictly by PR number per event type ensures rapid pushes cancel old setup jobs, and rapid suite completions cancel old notify calculations, but they don't cancel each other).*
2. **Explicit exit on no pull request:** The fallback `github.event.check_suite.head_sha` in the concurrency group handles fork PRs (where `pull_requests` array is empty). Since you document that fork PRs get no verdict, this is handled by your `if:` condition in `notify`. Clean.

## Overall Assessment
**APPROVE-WITH-CONCERN**. The mathematical logic of the patch-id and the structural self-healing of the check-count are brilliant and physically sound. The only flaw is a mechanical race condition in the Actions concurrency grouping where a slow background check finishing can snipe the tier-labeler out of the sky. Fix the concurrency group, and this is ready for production.

JamesPagetButler:
## Round-1 fixes applied — c554937 + e2fd090

Two changes since the reviewed head (33f90e87), so per this PR's own convention the round-1 reviews are stale and a delta round follows:

1. **(live-test finding, pre-review push c554937)** `workflow_run` trigger added: Actions-created suites never fire `check_suite:completed` to workflows (recursion guard) — on #510, the notifier's first audience, all 17 checks went green and it never woke. All ten repo workflows enumerated; `check_suite` retained for non-Actions apps; maintenance note in header.
2. **(Gemini BLOCKING, Red Team concurred)** concurrency split — improved over the proposed `event_name` keying: **job-level groups** (`pr-tier-label-<PR>` / `pr-notify-<PR>`) separate the axes exactly while keeping notify single-flight across both its event types (event_name keying would have let check_suite-notify and workflow_run-notify race the upsert).

Probe-6 correction acknowledged (classifiers agree; redundant globs → #507 sweep).

Delta round dispatching against the new head.

— qbp-oppenheimer session

JamesPagetButler:
## Red Team — Round-2 Delta Review (Sabine / Grothendieck / Knuth)

**Scope:** delta `33f90e87..e2fd090` on `.github/workflows/pr-check-status.yml` (the only file changed: +39/−6). Two commits: `c554937` (workflow_run trigger) and `e2fd090` (axis-split concurrency, the round-1 BLOCKING fix). Verdict structure: each probe assessed; regression surface diffed.

### Probe traffic-light

| # | Probe | Verdict | Evidence |
|---|-------|---------|----------|
| 1 | Axis separation resolves Gemini's BLOCKING scenario | ✅ PASS | tier-label group `pr-tier-label-N`; notify group `pr-notify-N`. Disjoint prefixes → an old-SHA notify completion can no longer cancel a new-SHA tier-label run. Different cancellation namespaces. |
| 1b | notify(check_suite) & notify(workflow_run) share ONE group | ✅ PASS | `check_suite.pull_requests[0].number \|\| workflow_run.pull_requests[0].number`: for a check_suite event the LHS is N and the workflow_run context is null → `pr-notify-N`; for a workflow_run event the check_suite context is null → coalesces to RHS = N → `pr-notify-N`. Identical string both ways → genuine single-flight across both event types. Confirmed improvement over the round-1 proposal of event_name keying (which would have split notify into two racing groups). |
| 2a | workflow_run.pull_requests[] populates for same-repo PRs | ✅ PASS | GitHub populates `workflow_run.pull_requests[]` when the triggering run's head branch is in the SAME repo. The file's own KNOWN LIMITATION (line 26–28) already scopes this to same-repo branches; fork PRs get no verdict (accepted, solo-maintainer). The job-`if` guards `pull_requests[0] != null` so a fork run no-ops rather than erroring. |
| 2b | 10-workflow enumeration complete vs `grep '^name:'` | ✅ PASS | Repo has 11 named workflows; the list contains all 10 non-self workflows and correctly OMITS "PR check-status notifier" itself (listing self would be dead config — recursion-guarded regardless). Exact match. |
| 2c | Maintenance-note risk (unlisted new workflow can't finalize verdict) | 🟡 NON-BLOCKING | Real residual risk: a later-finishing UNLISTED workflow = missed final event. The in-file MAINTENANCE note (lines 38–40) documents it. Acceptable for merge; recommend a #507-class follow-up CI lint that asserts every `name:` in `.github/workflows/*.yml` (minus self) appears in this `workflow_run.workflows` list. Filed-as-follow-up, not a merge blocker. |
| 3 | #510 failure timeline closed by new triggers | ✅ PASS | Pre-fix: Actions-created check_suites never re-trigger the workflow (GitHub recursion guard), so the last Actions check finishing was structurally invisible — the notifier only woke on CodeQL suites. Post-fix: each listed Actions workflow's `workflow_run:completed` wakes notify; the PENDING-gate (lines 147–151) holds until the per-SHA check-runs API shows zero incomplete, so whichever Actions workflow finishes last (e.g. 'Lean Foundations Gate' / lake build) fires the wake that finds all-complete and posts. The observed failure mode is closed. |
| 4 | No regression in unchanged logic | ✅ PASS | `git diff 33f90e87..HEAD` touches ONLY: trigger block, concurrency (workflow→job level), notify `if`, and the `PR`/`SHA` env coalesces. Self-filter (`SELF`/line 145), slurp `add // []`, printf bodies, patch-id freshness (merge-base + `patch-id --stable`) are byte-identical. No behavioural drift outside the trigger/concurrency/env surface. |

### Knuth (correctness) note
The PENDING-gate makes the design *idempotent under over-triggering*: it is harmless that BOTH a check_suite and a workflow_run can wake notify for the same SHA — every wake recomputes from the check-runs API and either no-ops (pending) or upserts the single marker comment. The single notify concurrency group then serialises any genuinely-concurrent wakes so the upsert can't double-post. Sound.

### Grothendieck (structure) note
The axis split is the correct abstraction: tier-label is keyed on (PR) over the `pull_request` event, notify is keyed on (PR) over the union of completion events. Two orthogonal cancellation lattices sharing only the PR index — exactly what round-1 demanded. No shared mutable cancellation state remains between the axes.

### Verdict: **APPROVE**

Round-1 BLOCKING finding is resolved at the structural level, not patched over. One non-blocking follow-up (2c: enumeration-drift CI lint, #507-class). No regressions.

JamesPagetButler:
## Gemini — Round-2 Delta Review (Furey / Feynman) — via real Gemini API (`gemini-3-pro-preview`, session review-20260604-230032)

Round-1 was APPROVE-WITH-CONCERN with ONE BLOCKING (cross-event concurrency: shared PR-keyed workflow-level group + cancel-in-progress let a slow OLD-SHA notify cancel a NEW-SHA tier-label run). Delta judged: `33f90e87..e2fd090` (commits c554937 workflow_run trigger, e2fd090 axis-split fix).

### Verdict Table

| Probe | Status | Note |
|---|---|---|
| 1. Axis separation | PASS | Round-1 BLOCKING is **fully resolved**. Job-level groups prevent `notify` from cancelling `tier-label`. The `\|\|` coalesce ensures `check_suite` and `workflow_run` share the identical `pr-notify-<PR>` group. |
| 2. workflow_run enumeration | CONCERN (non-blocking) | Same-repo `pull_requests[]` population works, but the hardcoded 10-workflow list is a fragile maintenance liability. |
| 3. #510 failure closure | PASS | `workflow_run:completed` correctly bypasses the Actions recursion guard, providing the missing wake when the final Actions check finishes. |
| 4. Double-post hazard | PASS | Shared `pr-notify` group + `cancel-in-progress` + 0-PENDING gate + marker-based upsert comprehensively eliminate any double-comment race. |

**Accuracy:** The architectural fix cleanly handles the GitHub API constraints — `check_suite` cannot observe Actions finishing (recursion guard), and `workflow_run` patches that blind spot. The concurrency design is technically sound and scopes cancellations precisely to (PR × task).

**Completeness:** The manual enumeration leaves an operational gap — if a future engineer adds a new CI pipeline and misses the MAINTENANCE note, that workflow will silently fail to trigger verdicts when it is the last to complete.

**Suggestion (non-blocking, #507-class follow-up):** Add a fast CI lint asserting the `workflow_run.workflows` array matches the extracted `name:` fields of all active `.github/workflows/*.yml` (minus self). Do not rely on the comment alone.

### Verdict: **APPROVE**

Round-1 BLOCKING successfully resolved; #510 structurally closed; concurrency matrix now fundamentally safe. Proceed with merging. One non-blocking follow-up (enumeration-drift lint).

PR #512, opened 2026-06-05
feat(cth): #509 unified vNext apply — 202 anchors, zero unreconciled streams (+ README canonical pointer + v5_24 baseline)

JamesPagetButler:
## Summary

The full #509 reconciliation, in three commits:

1. **`244c3b0`** — README canonical pointer fix (AC3) + v5_24 third-stream baseline tracked (md5-verified)
2. **`3e3860b`** — pre-v0.3 baseline moved to `baselines/` (exempt from top-level v0.3 schema lint, same pattern as `migrations/`)
3. **`9350aff`** — **the APPLY**: all three adjudicated batches landed on the canonical ledger in one deterministic pass

## The apply (commit 3)

`scripts/cth_509_apply.py` reads only the structured adjudication record (`analysis/509-apply-rulings.json`, encoding #509 comments 4627495414/4627533376/4627480904/4627520669/4627540950) + the Batch-B decisions JSON. **Canonical: 150 → 202 anchors. 0 schema violations. Tier domain [1,3] holds.**

| Batch | Adjudicator | Applied |
|---|---|---|
| A — v5.13 fold-in (24) | @qbp-oppenheimer | 24/24 intaken, 0 dropped; 1 include-as-killed (CAMB-rerun, EC-DEAD chain + revival path); 2 downgrades to marginal (Fmax-7 own-data tension); nuclear PROOF cluster → conjectural mapping; canonical side-effect: `PRED-TOV-limit-sqrt-7-over-3` → **falsified-as-global** + bump-peak regime pointer |
| B — provenance backfill (74) | @qbp-oppenheimer | 58 theory / 16 theory-external; **Convention D recorded as standing discriminator**; hurwitz-quat → external |
| C — v5_24 intake (28+55) | @qbp-oppenheimer + @cth-implementor | 19 NASCENT+substrate (each with the mandatory empty-Substrate rider); 3 NASCENT insights; 3 standard; **3 truth-in-labelling RELABELS** (PROOF→INSIGHT/DERIV, phantom terminations, id_history preserved); §3: 122 field adoptions, kill metadata inviolable (verified), INST-ckm tier-0 rejected; 16 REF-* → theory-external + citation-integrity semantics; 5 tier-4 **CLAMPED**→3 (napkin #514); Q27-TOV rider not triggered |

[...]

PR #517, opened 2026-06-05
[infra] notifier: exit-127 fix + API PR-resolution + failure visibility

JamesPagetButler:
## Summary

The notifier's first production outage, root-caused and triple-fixed. Impact of the outage: PR #512's red link-check got **no verdict comment**, silently breaking the "everything resolved → beekeeper reviews" contract — the beekeeper found it by reading raw check states, which is exactly the failure mode this workflow exists to prevent.

| Bug | Root cause | Fix |
|---|---|---|
| **exit-127 crash** at the first live ALL-GREEN (#515's suites, run 26992022640) | markdown backticks inside double-quoted bash = command substitution — bash tried to execute `reviewed-sha` as a program. `bash -n` can't catch it (valid syntax, wrong semantics); reviews couldn't catch it (the ALL-GREEN path only runs post-merge) | all FRESH strings via single-quoted `printf`; **bug-class audit script run across the whole workflow: CLEAN** |
| **silent skip on #512** | `if:` trusted `workflow_run.pull_requests[]`, which GitHub populates unreliably (empty for a real same-repo PR) | job fires on any suite/run completion; PR resolved via `commits/{sha}/pulls` API (payload as fast path); exits quietly only when the SHA genuinely heads no open PR |
| **silent failure mode itself** | a crashed notifier produced no signal anywhere a human looks | `if: failure()` step posts a warning comment on the PR naming the failed run — who-watches-the-watcher |

## Process note (for the record)

#511's own test plan deferred the live test to "the next PR after merge" — and the next PR demonstrated both bugs. Deferred-live-test-as-informal-intention is the same defect class as PATTERN-01; the failure-visibility step is the structural answer (the system now reports its own outages), and post-merge verification on the NEXT PR is a checklist item in this PR's plan, not an intention.

- [x] YAML valid; embedded bash `bash -n` clean
- [x] backtick-in-double-quote audit: CLEAN (script in PR description history)
- [ ] CI green on this PR
- [ ] Post-merge live test on the next active PR: verdict comment appears; failure step exercised by observation window

## CTH anchor impact
- [x] N/A — CI workflow only.

### DIFF
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -143,15 +143,29 @@ jobs:
 
           # PR resolution: payload first, API fallback (payload pull_requests[]
           # is unreliable — empty for real same-repo PRs; observed on #512).
+          # EXACT head-SHA match only (review #517 blocking #1): the endpoint
+          # returns every open PR whose branch CONTAINS the commit, so a loose
+          # `first` can post the verdict to the wrong PR in stacked branches.
+          # Verdicts are per-head — a SHA that heads no open PR gets none.
           if [ -z "${PR:-}" ]; then
             PR=$(gh api "repos/$REPO/commits/$SHA/pulls" \
-                 --jq '[.[] | select(.state == "open")] | (first // {}) | .number // empty')
+                 --jq '[.[] | select(.state == "open" and .head.sha == "'"$SHA"'")] | (first // {}) | .number // empty')
           fi
           if [ -z "${PR:-}" ]; then
-            echo "no open PR found for $SHA — not a PR head; nothing to do."
+            echo "no open PR headed by $SHA — nothing to do."
             exit 0
           fi
-          echo "resolved PR #$PR for $SHA"
+          # Stale-verdict guard (review #517 blocking #2): per-SHA concurrency
+          # means an old-SHA run is not cancelled by a new push; without this
+          # guard it could PATCH the marker comment with an outdated verdict
+          # AFTER the correct one. If the PR head has advanced past our event
+          # SHA, this run is history — exit.
+          HEADNOW=$(gh pr view "$PR" --repo "$REPO" --json headRefOid --jq .headRefOid)
+          if [ "$HEADNOW" != "$SHA" ]; then
+            echo "stale event: PR #$PR head is $HEADNOW, event SHA $SHA — exiting."
+            exit 0
+          fi
+          echo "resolved PR #$PR for $SHA (current head)"
           # review-completeness gates excluded from TECH-GREEN (they cannot be
           # green before reviews exist — the deadlock this design dissolves).
           # Single source of truth for the excluded list:
@@ -248,7 +262,10 @@ jobs:
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
           set -u
+          # exact head-SHA match, same rule as the main resolver (review #517)
           PR=$(gh api "repos/$REPO/commits/$SHA/pulls" \
-               --jq '[.[] | select(.state == "open")] | (first // {}) | .number // empty' || true)
+               --jq '[.[] | select(.state == "open" and .head.sha == "'"$SHA"'")] | (first // {}) | .number // empty' || true)
           [ -z "${PR:-}" ] && { echo "no PR to warn"; exit 0; }
-          gh pr comment "$PR" --repo "$REPO" --body "⚠️ **pr-check-status notifier ERRORED** on \\\`$SHA\\\` — no verdict was posted for this head. Check the run: $RUN_URL. Do not trust any older verdict comment above." || true
+          # single-quoted printf — no hand-escaped backticks anywhere (127 class)
+          WARN=$(printf '⚠️ **pr-check-status notifier ERRORED** on `%s` — no verdict was posted for this head. Check the run: %s. Do not trust any older verdict comment above.' "$SHA" "$RUN_URL")
+          gh pr comment "$PR" --repo "$REPO" --body "$WARN" || true
--- a/.github/workflows/pr-check-status.yml
+++ b/.github/workflows/pr-check-status.yml
@@ -113,9 +113,15 @@ jobs:
     # IMPORTANT: this job name is the notifier's own check-run name; the query
     # below excludes it by name (self-counting fix, #508 review finding #1).
     name: pr-check-status-notify
-    if: ${{ (github.event_name == 'check_suite' && github.event.check_suite.pull_requests[0] != null) || (github.event_name == 'workflow_run' && github.event.workflow_run.pull_requests[0] != null) }}
+    # Fire on any suite/run completion that carries a head SHA; the PR is
+    # resolved via API inside the job. The event payload's pull_requests[] is
+    # NOT trusted: GitHub populates it unreliably (observed empty for real
+    # same-repo PR #512 → the notifier silently skipped a red verdict).
+    if: ${{ github.event_name == 'check_suite' || github.event_name == 'workflow_run' }}
     concurrency:
-      group: pr-notify-${{ github.event.check_suite.pull_requests[0].number || github.event.workflow_run.pull_requests[0].number }}
+      # group by SHA (PR number unavailable pre-resolution); same-PR pushes get
+      # fresh groups per SHA — acceptable, the upsert is idempotent per comment
+      group: pr-notify-${{ github.event.check_suite.head_sha || github.event.workflow_run.head_sha }}
       cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
@@ -134,6 +140,18 @@ jobs:
           set -euo pipefail
           MARKER="<!-- pr-check-status-notifier -->"
           SELF="pr-check-status-notify"
+
+          # PR resolution: payload first, API fallback (payload pull_requests[]
+          # is unreliable — empty for real same-repo PRs; observed on #512).
+          if [ -z "${PR:-}" ]; then
+            PR=$(gh api "repos/$REPO/commits/$SHA/pulls" \
+                 --jq '[.[] | select(.state == "open")] | (first // {}) | .number // empty')
+          fi
+          if [ -z "${PR:-}" ]; then
+            echo "no open PR found for $SHA — not a PR head; nothing to do."
+            exit 0
+          fi
+          echo "resolved PR #$PR for $SHA"
           # review-completeness gates excluded from TECH-GREEN (they cannot be
           # green before reviews exist — the deadlock this design dissolves).
           # Single source of truth for the excluded list:
@@ -179,10 +197,14 @@ jobs:
             LASTREV=$(gh api "repos/$REPO/issues/$PR/comments?per_page=100" --paginate \
                       --jq '[.[] | .body]' | jq -s 'add // []' \
                       | jq -r '[.[] | capture("<!-- reviewed-sha: (?<s>[0-9a-f]{7,40}) -->").s] | last // empty')
-            FRESH="⚠️ review freshness UNVERIFIABLE — no `reviewed-sha` marker found in review comments; confirm reviews cover this head before merging."
+            # NOTE: FRESH strings are built with single-quoted printf formats
+            # ONLY — markdown backticks inside double-quoted bash are command
+            # substitution (the exit-127 incident: bash tried to EXECUTE
+            # `reviewed-sha` at the notifier's first live ALL-GREEN).
+            FRESH=$(printf '⚠️ review freshness UNVERIFIABLE — no `reviewed-sha` marker found in review comments; confirm reviews cover this head before merging.')
             if [ -n "$LASTREV" ]; then
               if [ "$LASTREV" = "$SHA" ]; then
-                FRESH="✔️ reviews stamped for this exact head."
+                FRESH=$(printf '✔️ reviews stamped for this exact head.')
               else
                 # patch-id comparison: identical content (rebase / empty
                 # retrigger commits) keeps reviews fresh; content change
@@ -194,9 +216,9 @@ jobs:
                 P_HEAD=$(pid "$SHA" || echo head-err)
                 P_REV=$(pid "$LASTREV" || echo rev-err)
                 if [ -n "$P_HEAD" ] && [ "$P_HEAD" = "$P_REV" ]; then
-                  FRESH="✔️ reviews stamped for \`$LASTREV\`; patch-id identical to head (cosmetic motion only — rebase/retrigger). Reviews remain fresh."
+                  FRESH=$(printf '✔️ reviews stamped for `%s`; patch-id identical to head (cosmetic motion only — rebase/retrigger). Reviews remain fresh.' "$LASTREV")
                 else
-                  FRESH="🔶 **DELTA ROUND REQUIRED** — reviews stamped for \`$LASTREV\` but the diff content has changed since (patch-id mismatch). Do not merge on these reviews."
+                  FRESH=$(printf '🔶 **DELTA ROUND REQUIRED** — reviews stamped for `%s` but the diff content has changed since (patch-id mismatch). Do not merge on these reviews.' "$LASTREV")
                 fi
               fi
             fi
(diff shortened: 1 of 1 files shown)


JamesPagetButler:
## 🔴 Red Team Review — Round 3 (Knuth lead) — PR #517

> Third review round on this single file. Its track record: **four** structural bugs caught across its life — self-counting (#508 #1), pagination one-array-per-page, check_suite Actions-recursion-guard blindness, and now exit-127 + payload-skip. Reviewed with that humility: a file that has fooled `bash -n`, fooled prior reviews, and only revealed bugs at *live* ALL-GREEN warrants byte-level paranoia, not pattern-matching reassurance.

### Verdict table

| Probe | Finding | Severity |
|---|---|---|
| 1. exit-127 completeness | No recurrence in success paths (printf single-quote, verified). **Failure step renders `\`sha\`` with visible backslashes** — cosmetic, not a crash. | NON-BLOCKING (cosmetic) |
| 2. `commits/{sha}/pulls` jq | **No `select(.head.sha==$SHA)` filter** — `first` can pick a base/stacked PR that merely *contains* the commit. Wrong-PR verdict risk. | **BLOCKING** |
| 3. per-SHA concurrency | **Stale-SHA run can PATCH the marker with an outdated verdict after the new SHA posted.** Old #511 per-PR keying killed exactly this; the regression is real. | **BLOCKING** |
| 4. `failure()` on cancellation | Correct — `failure()` excludes cancellation; concurrency-cancelled jobs skip the warning step. No spam. | PASS |
| 5. repo-wide `if:` side-effects | Clean — `add_label`/`remove_label` only *called* after the PR-resolution guard; exit-0 path touches nothing. | PASS |

---

### Probe 1 — the 127 fix, walked to the bytes

Independently re-ran a backtick-in-double-quote audit over every `run:` block. The success-path FRESH/BODY strings are all single-quoted `printf` — **clean**, the fix is correct and complete for those.

The **NEW failure step** (line 254) is the interesting one. It uses `\\\`$SHA\\\`` inside a double-quoted `--body`. Walked character by character:

- YAML literal block scalar (`run: |`) does **not** process backslash escapes — the bytes reach bash verbatim: `\\\`$SHA\\\``.
- In a bash double-quoted string: `\\` → one literal `\`; `\`` → one literal backtick. So bash produces `` \`<sha>\` `` — a backslash followed by a backtick, twice.
- **No command substitution fires** (the backtick is escaped). So this is **NOT** an exit-127 recurrence. Good — the fix held.
- **But** the rendered markdown is `` \`abc123\` `` → GitHub renders backslash-escaped backticks as *literal* backtick characters, so the beekeeper sees `` `abc123` `` *with visible backslashes*, not a code span. In the one message whose entire job is to be readable during an outage, that's a blemish.

Empirical proof:
```
SUCCESS-path body : [on `abc123` done]
FAILURE-path body : [on \`abc123\` — done]
```

Fix: drop to the same single-quoted `printf` pattern the rest of the file standardized on —
```bash
gh pr comment "$PR" --repo "$REPO" --body \
  "$(printf '⚠️ **pr-check-status notifier ERRORED** on `%s` — no verdict was posted for this head. Check the run: %s. Do not trust any older verdict comment above.' "$SHA" "$RUN_URL")" || true
```
This both fixes the rendering *and* removes the last hand-escaped backtick in the file — closing the bug class entirely rather than dodging it once more. NON-BLOCKING but strongly recommended: the file's whole history is "the backtick path we didn't standardize bit us."

### Probe 2 — `commits/{sha}/pulls` returns the wrong PR for stacked branches *(BLOCKING)*

The endpoint is the correct, maintained replacement for the deprecated `application/vnd.github.groot-preview` API, and it **does** reliably list same-repo-branch PRs (this is the right fix for the #512 payload-skip — confirmed). 

But the endpoint returns **every open PR whose branch contains `$SHA`** — not only the PR whose *head* is `$SHA`. Two ways this misfires:

1. **Stacked PRs:** PR-A (feature) and PR-B (stacked on A). A commit on A's tip is contained by both branches → endpoint returns both → `(first // {}) | .number` picks one **arbitrarily by API ordering**. The verdict can land on the wrong PR.
2. The current jq filters only `select(.state=="open")`, never `select(.head.sha == $SHA)`.

The notifier's contract is "verdict for *this head SHA*." Resolve the PR **whose head is this SHA**, preferring an exact head match:
```bash
PR=$(gh api "repos/$REPO/commits/$SHA/pulls" \
     --jq --arg sha "$SHA" '[.[] | select(.state=="open")]
            | ( map(select(.head.sha == $sha)) + . )
            | (first // {}) | .number // empty')
```
(The `map(head.sha==$sha) + .` prefers an exact-head match, falls back to any-open if none — so a mid-flight push that already advanced head still resolves, but a stacked pair resolves to the right PR.) Same fix needed in the **failure step** (line 251), which has the identical unfiltered query.

### Probe 3 — per-SHA concurrency lets a stale verdict overwrite a fresh one *(BLOCKING)*

The group changed from `pr-notify-<PR#>` (#511) to `pr-notify-<head_sha>`. Walk the race:

1. Push SHA-1 to PR. Suite-A completes → notify run R1 (group `pr-notify-SHA1`) starts, slow (patch-id fetch, paginated comment scan).
2. Beekeeper pushes SHA-2. Suite-B completes → notify run R2 (group `pr-notify-SHA2`) starts. **Different group → R2 does NOT cancel R1.**
3. R2 finishes first, PATCHes the marker comment: "ALL GREEN @ SHA-2."
4. R1 (still running on SHA-1's stale check set) finishes and PATCHes the **same** marker comment: "Checks FAILED @ SHA-1" (or a stale ALL-GREEN).

Result: the canonical verdict comment now shows an **outdated verdict for a superseded SHA**, posted *after* the correct one. This is precisely the failure mode #511's per-PR keying (`cancel-in-progress` within one PR group) was designed to kill — moving the key to SHA reintroduced it. The PR body's "the upsert is idempotent per comment" is true but irrelevant: idempotency guarantees one comment, not the *latest* verdict in it.

The header comment (lines 66–73) justifies job-level groups to separate notify/tier-label axes — sound — but conflates that with the *keying granularity*. You can keep job-level separation and still key notify per-PR. The problem: PR# isn't known pre-resolution, and you can't re-key `concurrency:` after a step resolves it.

Two viable fixes:
- **(A) Tiny resolver job + per-PR notify group.** A `resolve` job (fires on suite/run completion) computes PR# from `$SHA` via the same API and exposes it as an output; `notify` depends on it with `concurrency: group: pr-notify-${{ needs.resolve.outputs.pr }}`. Concurrency expressions *can* reference `needs` outputs. This restores exact per-PR cancellation.
- **(B) Stale-guard in the body step.** Before composing, fetch current PR head and bail if it moved:
  ```bash
  CUR=$(gh api "repos/$REPO/pulls/$PR" --jq .head.sha)
  [ "$CUR" != "$SHA" ] && { echo "head advanced $SHA→$CUR; superseded, skipping"; exit 0; }
  ```
  Cheaper, no job restructure; the race window shrinks to between the check and the PATCH (sub-second, and the loser exits clean). (A) is airtight; (B) is 90% with three lines. Either resolves BLOCKING.

### Probe 4 — failure step does not spam cancelled runs *(PASS)*
`failure()` is true only if a prior step *failed*; a concurrency-cancelled job is `cancelled()`, not `failed()`, so the warning step is skipped on cancellation. No noise. Correct as written.

### Probe 5 — repo-wide `if:` has no side-effects before the guard *(PASS)*
The job now fires for every suite/workflow_run completion repo-wide (incl. master pushes). The exit-0 path runs: checkout → resolve PR → `exit 0` on no-PR. `add_label`/`remove_label` are only *defined* before the guard (functions, no-op) and only *invoked* in the RED/TECH-GREEN branches well after the guard. No comment, no label touched before PR resolution. Cost is seconds of a runner per event — acceptable. Clean.

---

Linked work: the PR body's three-bug table is the de-facto AC (no separate issue number in title/body).

| Criterion (from PR body) | Assessment | Evidence |
|---|---|---|
| exit-127 fixed; bug-class audit CLEAN | **PARTIAL** | Success paths clean; failure step still hand-escapes backticks → cosmetic render defect (Probe 1). The "CLEAN" claim is true for execution, not for the last backtick instance. |
| payload-skip fixed via `commits/{sha}/pulls` | **PARTIAL** | Endpoint choice correct; resolution unfiltered by head.sha → wrong-PR risk (Probe 2). |
| failure-visibility step posts warning | **PASS (functional)** | Step present, fires on `failure()`, resolves PR, posts. Renders with backslash blemish (Probe 1). |

**Verdict: REQUEST-CHANGES.** Two BLOCKING (Probe 2 wrong-PR resolution; Probe 3 stale-verdict concurrency race) + one recommended cosmetic (Probe 1). The functional fixes are real and the diagnosis of both production bugs is correct — but the file's history is "structural bug survived because it only manifests live," and Probes 2 & 3 are exactly that shape: invisible until a stacked PR or a fast double-push hits them. Hold for the head.sha jq filter (both queries) and a concurrency stale-guard before merge.

— Knuth (Red Team), with Sabine (race/payload empiricism) and Grothendieck (the resolution-correctness invariant: *resolve the PR whose head IS this SHA*, not one that merely contains it)

JamesPagetButler:
## 🔵 Gemini Review (gemini-3-pro-preview, real API) — Round 3 — PR #517

Sequential review, posted after Red Team. Personas: Furey (structural invariants) + Feynman (operational reality). Session `review-20260605-144030`.

---

### Clarity
The two-phase review trigger model (Tech-Green vs All-Green) is well-defined. The bash logic for evaluating checks and formatting markdown is dense but appropriately centralized. Using `patch-id` to detect cosmetic rebases is a brilliant, physics-like invariant check — it measures the "true diff" invariant rather than relying on the coordinate SHA.

### Accuracy
I independently verified Red Team's findings and Probes 4 & 5.

- **BLOCKING A (Probe 2 — Stacked PR mismatch): CONCUR.** The `commits/$SHA/pulls` endpoint returns all PRs that contain the commit. Relying on `first` is non-deterministic for stacked branches and can post the verdict to the wrong PR. Adding the `head.sha` filter is structurally necessary.
- **BLOCKING B (Probe 3 — Concurrency Race): CONCUR.** Shifting concurrency to the SHA means older and newer runs for the same PR execute in parallel. The older run can finish slower and overwrite the single PR marker comment with a stale verdict. Red Team's stale-guard suggestion (`CURRENT_HEAD != $SHA`) is the most elegant, minimalist fix.
- **NON-BLOCKING (Probe 1 — Backslash escaping): CONCUR.** `\\\`$SHA\\\`` inside a double-quoted bash string resolves to `\`$SHA\``, which markdown renders as a literal backtick preceded by a visible backslash, breaking the code span.
- **Probes 4 & 5 (re-verify): CONCUR WITH PASS.** `failure()` ignores `cancelled` runs, preventing warning-spam. The PR-resolution guard executes before any mutating `gh pr edit` or `POST`, so the exit-0 path is side-effect free.

**NEW (Gemini) — jq `capture` crash claim:** Gemini asserted the `LASTREV` line crashes because `capture()` errors on non-matching strings.

**NEW (Gemini) — checkout syntax anomaly:** Gemini reported a malformed `uses: actions/checkout@...json`.

### Suggestions (Gemini)
1. **Fix BLOCKING A:** `--jq '[.[] | select(.state == "open" and .head.sha == "'"$SHA"'")] | (first // {}) | .number // empty'`
2. **Fix BLOCKING B:** stale-guard after resolution:
   ```bash
   CURRENT_HEAD=$(gh pr view "$PR" --repo "$REPO" --json headRefOid -q .headRefOid)
   if [ "$CURRENT_HEAD" != "$SHA" ]; then
     echo "SHA $SHA is no longer head; exiting to prevent stale overwrites."; exit 0
   fi
   ```
3. ~~Fix jq crash with `?`~~ — see Herschel verification below.
4. **Fix NON-BLOCKING:** single-quoted printf for the failure body.
5. ~~Revert checkout to @v4~~ — see Herschel verification below.

### Gemini overall: **REQUEST-CHANGES.**

---

## 🔎 Herschel verification of Gemini's two NEW claims (Red Team, before propagating)

Per workflow rule, a review's new BLOCKING findings are verified before they gate a merge. Both Gemini NEW findings were tested empirically against the actual file and `jq`:

| Gemini NEW finding | Verdict | Evidence |
|---|---|---|
| jq `capture()` crashes on non-matching comments → guaranteed runtime crash | **FALSE POSITIVE — withdrawn** | `capture()` returns `null` (not an error) on no-match in jq 1.5/1.6/1.7. Ran the **exact** workflow pipeline on a mixed array (matching + non-matching + zero-match) — exit 0, correct SHA extracted, no crash. The `?` operator is unnecessary. |
| `uses: actions/checkout@…json` malformed | **HALLUCINATION — does not exist** | The file at the reviewed SHA reads `uses: actions/checkout@v4` (line 130). No such malformed string is present; Gemini fabricated it. |

Empirical proof of the jq claim:
```
$ echo '["random text","<!-- reviewed-sha: fce52ac... -->","more text"]' \
   | jq -r '[.[] | capture("<!-- reviewed-sha: (?<s>[0-9a-f]{7,40}) -->").s] | last // empty'
fce52acbb02d6e172260968f6ed4ff3377bdb2c8      # exit 0, no crash
$ echo '["foo","bar"]' | jq -r '[.[] | capture(...).s] | last // empty'
                                               # empty, exit 0, no crash
```

**Net: Gemini's two NEW findings do not survive verification and do NOT gate this PR.** The two findings Gemini *concurs with Red Team on* (Probe 2 wrong-PR resolution, Probe 3 stale-verdict concurrency race) DO stand and are the real BLOCKING items. Gemini's verdict direction (REQUEST-CHANGES) is correct; the reasoning is sound on the concurred findings and incorrect on the two it originated.

JamesPagetButler:
## Red Team Delta Review — Round 2 (APPROVE)

**Personas:** Sabine (empirical skepticism), Grothendieck (structural correctness), Knuth (exactness)
**Scope:** delta only — fix commit `9311433` against round-1 blockers. Verified against the working-tree file at head, plus live shell/jq tests.

### Round-1 blockers — resolution status

| # | Round-1 blocker | Fix | Verified | Verdict |
|---|---|---|---|---|
| 1 | Wrong-PR resolution: loose `first` could post to a PR that merely *contains* the commit (stacked branches) | Both resolver sites now `select(.state == "open" and .head.sha == "'"\$SHA"'")`; loose fallback removed | jq filter tested live: 2 open PRs (one heads SHA, one contains a different SHA) + closed PR sharing SHA → returns **517 only**; SHA heading no open PR → **empty** | **RESOLVED** |
| 2 | Stale-verdict race: per-SHA concurrency leaves old-SHA runs uncancelled; could PATCH an outdated verdict after the correct one | Stale-guard `gh pr view --json headRefOid`; `exit 0` if head advanced | Guard at L164 executes **before** first label mutation (L200/206) and comment PATCH/POST (L247/250) | **RESOLVED** |
| 3 (cosmetic) | Hand-escaped backticks (127 class) | Failure-step WARN via single-quoted printf | Backtick audit CLEAN — every backtick is inside a single-quoted printf format or a YAML comment; none inside a double-quoted bash string | **RESOLVED** |

### Verification checklist (all PASS)

- **(a) Quoting walk of `'"\$SHA"'`** — bash concatenation `'...== "'"\$SHA"'")...'` renders the jq program literal `... .head.sha == "<sha>")`, a correctly quoted JSON string literal. Tested live; exact-match selection correct. **PASS**
- **(b) Stale-guard ordering** — guard precedes ALL side-effects (labels + comment upsert). **PASS**
- **(c) Failure-step printf** — renders a proper markdown code span, zero backslashes (tested). ALL-GREEN body with embedded `beekeeper's call` apostrophe renders cleanly. **PASS**
- **(d) No regression** — `failure()` cancellation semantics intact; exit-0 side-effect-free path preserved (both `exit 0` before any mutation); 127-class audit re-run = CLEAN; YAML parses OK. **PASS**
- **(e) Exact-match-only interplay** — when a suite completes for an OLD SHA after a push, the resolver finds no open PR heading that SHA → quiet `exit 0` at "no open PR headed by $SHA". This is the *desired* belt-and-braces: the resolver early-exit and the stale-guard both close the stale window with no gap. No verdict that *should* post for the current head is lost. **PASS**

### Verdict: **APPROVE**

No blocking items. Both round-1 blockers resolved and independently re-verified; no regressions to round-1-passed items.

JamesPagetButler:
## Gemini Delta Review — Round 2 (APPROVE)

**Personas:** Furey (structural/algebraic rigor), Feynman ("does it actually work")
**Engine:** real Gemini API, model `gemini-3-pro-preview`, session `review-20260605-144726`
**Method:** verbatim inlined hunks from head `9311433`; instructed to cite inlined text only (round-1 had two empirically-disproven blockers: a `jq capture()` null-crash claim and a hallucinated `actions/checkout` string).

### Findings

| Item | Gemini assessment | Cited lines |
|---|---|---|
| Blocking #1 (wrong-PR) | **Fixed.** Both jq filters select exact head match; loose fallback gone — `first` now acts only on the strictly filtered array | ~152 (main), ~267 (failure step) |
| Blocking #2 (stale-verdict) | **Fixed.** Stale-guard executes before all side-effects (labels 200/206, comments 247/250) | 163–167 |
| New risk (exact-match-only) | **Structurally sound.** Old-SHA completion fails the exact-match filter and exits quietly; the `gh pr view` check is belt-and-braces. No current-head verdicts lost | resolver + 163–167 |
| Regression check | None. Failure path exact-match + printf align with main path; no new edge-case regressions | — |

### Verdict: **APPROVE**

> "The fixes correctly address the PR resolution bug and the stale-verdict race condition with robust, side-effect-free early exits."

No blocking items.

Issue #525, opened 2026-06-08
[foundations] exp/log enhancements: full-ℝ one-parameter group law + left-inverse on the principal strip


JamesPagetButler:
Follow-up from PR #522 (PR-F exp/log) Tier-3 review. The merged exp/log is rigorous and the #474 exp/log CELL is met (exp total + well-defined; log on the dense open domain {N x>0 ∧ Im x≠0} with exp(log x)=x proven; n=2 Mathlib-quaternion-exp agreement proven). These are BEYOND-CELL enhancements, tracked here:

1. **Full-ℝ one-parameter group law.** `exp_smul_add` currently carries s,t ≥ 0 (nonnegative-cone sub-monoid law; the proof uses Real.sqrt_sq on imNorm(q•x)=q‖Im x‖, valid only q≥0). Extend to all s,t ∈ ℝ.
2. **Left-inverse** `log(exp x) = x` on the principal strip (imNorm < π) — currently only the right-inverse exp(log x)=x is proven.

**AC:**
- [x] exp((s+t)•x) = exp(s•x)·exp(t•x) for all s,t : ℝ
- [x] log(exp x) = x on the principal strip
- [x] zero sorry / native_decide; axioms ⊆ standard triple

---
## On close: required upgrade (close-your-loops + pending-upgrade guard)
This issue is the unblocking item for **#474 AC6 exp/log = MET-pending** (register: `docs/foundations/pending-upgrades.md`). The `check_pending_upgrades.py` CI guard (#526) goes RED when this issue closes while its register row is still `status: pending`. So closing this issue REQUIRES, in the same change:
- [x] flip #474 AC6 exp/log cell pending→fully-met (AC tracking + the matrix index annotation, PR-G)
- [x] mark the register row `status: resolved` in `docs/foundations/pending-upgrades.md`

JamesPagetButler:
## Record correction — ACs ticked (all verified met)
#525 auto-closed via #544's `Closes #525` on merge, but its AC boxes were left unticked (the merge predated the FAULT-S4-004 close-guard). The work is genuinely complete on master — ticking for an accurate record:
- exp_smul_add_real (full-ℝ group law) ✓ · log_exp (principal-strip left-inverse) ✓
- zero sorry/native; axioms ⊆ {propext, Classical.choice, Quot.sound} ✓ (independently re-verified)
- #474 AC6 flipped pending→full ✓ · pending-upgrades register `status: resolved` ✓
Correctly closed; record now matches reality.

PR #522, opened 2026-06-08
[foundations] #474 PR-F: full-rigor exp/log via quadraticity

JamesPagetButler:
## Summary

Wave-2 PR-F (issue #474) — the analytic row, at FULL rigor per the beekeeper ruling, with no non-associative ambient power series.

Route: every x ∈ CDAlg ℝ n has x² = 2 Re(x)·x − N(x)·1 (merged `cdAlg_sq_eq`), so span{1, Im x} ≅ ℂ; exp defined closed-form there. Proven: `exp_zero`, the **nonnegative-cone** exponential law `exp((s+t)•x)=exp(s•x)*exp(t•x)` for **s,t ≥ 0** (`octonion_exp_smul_add`; sub-monoid law, NOT yet the full ℝ one-parameter group — see follow-up), N(exp x) = exp(2 Re x), exp(−x)·exp x = 1, and a principal `log` with proven right-inverse (`sedenion_exp_log`). 𝕆/𝕊 cells are the n=3,4 instances.

## Provenance note
Authored pre-crash; recovered, re-verified (build 3119 jobs + gate + axioms), committed post-crash. Not taken on the stalled agent's word. NOTE for review: the agent stalled before its self-report, so the exact scope of `log`'s proven domain and n=2 Mathlib-exp agreement is PROVEN (exp_eq_quaternion) — confirmed by reviewers against the file (honest-gap audit) rather than a delivery summary.

## Theory-element headline
N/A — pure-math foundation; no physical claim.

## CTH anchor impact
- **New anchors:** none.
- **Anchor-rule terminations:** [x] Lean file:line — Exp.lean
- **Routing axis** (per `docs/workflows/pr7_conflict_routing_rubric.md` v0.2):
  - [x] theory-axis → co-sign by @qbp-oppenheimer needed (default for Tier 3)

## Mechanical verification evidence
- [x] `lake build QBP.Foundations.Exp` green (3119 jobs)
- [x] Foundations gate PASSED 0/0; layer imports clean; aggregator wired
- [x] `#print axioms`: all ⊆ {propext, Classical.choice, Quot.sound}
- [x] Zero sorry / native_decide / vacuous-True
- [x] CI green — 14/14 checks confirmed mergeable @ 2026-06-08 (notifier verdict lagged; checks re-queried green)

Part of #474 (Wave 2, AC6 — exp/log).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #474, opened 2026-06-01
[foundations] [PHASE-1] Operations-complete tower ℝ→ℂ→ℍ→𝕆→𝕊 (proof-coverage manifest)


JamesPagetButler:
## Summary

Tracking issue for the Phase-1 operations-complete tower: prove ℝ→ℂ→ℍ→𝕆→𝕊 zero-sorry, with every cell of the property matrix either a real theorem (✓) or a concrete Lean counterexample term in Breakdown.lean (✗). This is the exit gate for Phase 1 of the foundations rebuild.

**Provenance:** ratified by beekeeper 2026-05-31; see `docs/foundations/scope-deliberation-2026-05-31.md` §3 (property matrix) and §2 (phased plan); sessionbridge `pr407-conflict-resolution` seq 69-74.

---

## Property Matrix (Phase-1 exit spec)

Every cell is a proof obligation. ✓ = theorem; ✗ = **concrete Lean counterexample term in `Breakdown.lean`** (via `decide` or explicit construction — NEVER `sorry`, NEVER `True := by trivial`). This anti-pattern hid the `octonionMul` defect (#472) and is the architecture-mandated exit gate.

| Operation / Law | ℝ | ℂ | ℍ | 𝕆 | 𝕊 |
|---|:--:|:--:|:--:|:--:|:--:|
| Addition, 0, −x, ℝ-scaling | ✓ | ✓ | ✓ | ✓ | ✓ |
| Multiplication, 1 | ✓ | ✓ | ✓ | ✓ | ✓ |
| Conjugation x* (anti-automorphism) | ✓ | ✓ | ✓ | ✓ | ✓ |
| Norm N(x)=xx*, Re/Im parts | ✓ | ✓ | ✓ | ✓ | ✓ |
| Bilinear/inner-product form ⟨x,y⟩=Re(x·ȳ) | ✓ | ✓ | ✓ | ✓ | ✓ |
| exp / log (well-defined via power-associativity) | ✓ | ✓ | ✓ | ✓ | ✓ |
| Cross product on Im (3D at ℍ, 7D at 𝕆) | — | — | ✓ | ✓ | — |
| Cayley-Dickson doubling (constructor) | — | ✓ | ✓ | ✓ | ✓ |
| **Total order** | ✓ | **✗** | **✗** | **✗** | **✗** |
| **Commutativity** | ✓ | ✓ | **✗** | **✗** | **✗** |
[...]

PR #526, opened 2026-06-08
[infra] pending-upgrade guard — closed unblocking issue forces the owed upgrade

JamesPagetButler:
## Summary

The mechanical path for **"AC met-pending-#NNN"** rulings (#522's AC6 exp/log, ruled met-pending-#525). Makes the deferred upgrade un-skippable.

| Piece | What |
|---|---|
| `docs/foundations/pending-upgrades.md` | register — one row per conditionally-complete item → unblocking issue + `status: pending\|resolved` |
| `scripts/check_pending_upgrades.py` | CI guard — RED when a pending row's unblocking issue is CLOSED; skips gracefully if `gh` unavailable |
| `lean-foundations.yml` | runs the guard with `GH_TOKEN` + trigger paths |

The moment #525 merges, this goes red until AC6 is flipped pending→full and the row marked `resolved` — promise-to-update becomes a merge-gate (PATTERN-01: only an un-mergeable state holds). First entry: #474 AC6 → #525.

## Theory-element headline
N/A — CI tooling; no physical claim.

## CTH anchor impact
- **New anchors:** none.
- **Routing axis**:
  - [x] N/A — process tooling, no anchor-bearing paths

- [x] guard exits 0 now (#525 OPEN); would exit 1 on a closed unblocking issue
- [x] YAML valid; black/mypy clean
- [x] CI green — all checks pass + mergeable; Red Team + Gemini both APPROVE (probe-5 blocker confirmed resolved) @ 2026-06-08

Refs #507, #474/#522/#525.

### DIFF
--- a/.github/workflows/lean-foundations.yml
+++ b/.github/workflows/lean-foundations.yml
@@ -23,8 +23,6 @@ on:
       - 'proofs/QBP/Foundations/.lean-gate-baseline.json'
       - 'scripts/check_lean_foundations.py'
       - 'scripts/check_layer_imports.py'
-      - 'scripts/check_pending_upgrades.py'
-      - 'docs/foundations/pending-upgrades.md'
       - '.github/workflows/lean-foundations.yml'
 
 permissions:
@@ -43,10 +41,6 @@ jobs:
         run: python3 scripts/check_lean_foundations.py --dir proofs/QBP/Foundations
       - name: Enforce layer-import DAG + aggregator completeness
         run: python3 scripts/check_layer_imports.py
-      - name: Pending-upgrade guard (closed unblocking issue ⇒ owed upgrade must land)
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: python3 scripts/check_pending_upgrades.py
 
   lake-build:
     name: lake build (zero errors)
--- a/.github/workflows/lean-foundations.yml
+++ b/.github/workflows/lean-foundations.yml
@@ -23,6 +23,8 @@ on:
       - 'proofs/QBP/Foundations/.lean-gate-baseline.json'
       - 'scripts/check_lean_foundations.py'
       - 'scripts/check_layer_imports.py'
+      - 'scripts/check_pending_upgrades.py'
+      - 'docs/foundations/pending-upgrades.md'
       - '.github/workflows/lean-foundations.yml'
 
 permissions:
@@ -41,6 +43,10 @@ jobs:
         run: python3 scripts/check_lean_foundations.py --dir proofs/QBP/Foundations
       - name: Enforce layer-import DAG + aggregator completeness
         run: python3 scripts/check_layer_imports.py
+      - name: Pending-upgrade guard (closed unblocking issue ⇒ owed upgrade must land)
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: python3 scripts/check_pending_upgrades.py
 
   lake-build:
     name: lake build (zero errors)
--- a/.github/workflows/pending-upgrades.yml
+++ b/.github/workflows/pending-upgrades.yml
@@ -1,0 +1,30 @@
+# Pending-upgrade guard — runs on EVERY PR (review #526 probe-6).
+#
+# The guard must trip on ANY pull request once a registered unblocking issue
+# closes — not only on foundations-touching PRs. A foundations-scoped trigger
+# would let the owed upgrade be skipped by any non-foundations PR merging first,
+# undercutting the "un-skippable the moment #NNN closes" guarantee. So it lives
+# in its own all-paths workflow.
+name: Pending-upgrade guard
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+  issues: read        # gh issue view — explicit (review #526 probe-7: survives
+                      # the repo ever going private; not inherited from contents)
+
+jobs:
+  pending-upgrades:
+    name: Pending-upgrade guard (closed unblocking issue ⇒ owed upgrade must land)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Enforce pending-upgrade register
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: python3 scripts/check_pending_upgrades.py
--- a/docs/foundations/pending-upgrades.md
+++ b/docs/foundations/pending-upgrades.md
@@ -1,0 +1,25 @@
+# Pending Upgrades Register
+
+Tracks deliverables that are **conditionally complete** — an acceptance-criterion
+cell or claim ruled MET for now, but with a known enhancement owed once a named
+unblocking issue lands. The companion CI guard (`scripts/check_pending_upgrades.py`)
+**fails the build when an unblocking issue here is CLOSED but its row is still
+`status: pending`** — converting "we'll update it later" from a memory-dependent
+promise into a mechanical merge-gate. (PATTERN-01 lesson: promises-to-update rot;
+only an un-mergeable state holds.)
+
+## How to resolve a row when its unblocking issue closes
+
+1. Apply the upgrade the row describes (the AC cell / annotation / doc edit).
+2. Change the row's `status:` from `pending` to `resolved` (keep the row as a
+   forensic record — mark, don't delete), or delete the row if the register entry
+   itself was the only artifact.
+3. CI goes green again.
+
+## Register
+
+<!-- one row per pending upgrade; the guard parses: issue: #N | status: pending|resolved -->
+
+| item | ruled | upgrade owed | unblocking issue | status |
+|---|---|---|---|---|
+| #474 AC6 exp/log cell | MET-pending (beekeeper, 2026-06-08) — exp total + log on dense domain + exp∘log=id clears "well-defined"; full-ℝ group law & left-inverse are enhancements | flip AC6 exp/log cell pending→full in #474 tracking + matrix index (PR-G) annotation | issue: #525 | status: pending |
--- a/scripts/check_pending_upgrades.py
+++ b/scripts/check_pending_upgrades.py
@@ -33,8 +33,14 @@ REGISTER = (
 ROW_RE = re.compile(r"issue:\s*#(\d+)\s*\|\s*status:\s*(pending|resolved)", re.I)
 
 
-def gh_issue_state(num: str) -> str | None:
-    """OPEN / CLOSED, or None if gh can't answer."""
+def gh_issue_state(num: str) -> str:
+    """OPEN / CLOSED / NOT_FOUND / TRANSIENT.
+
+    Distinguishes a genuinely-missing issue (a typo'd register number — must be
+    a loud violation, else the row silently never guards: the exact silent-rot
+    this whole guard exists to kill, review #526 probe-5) from a transient gh
+    failure (network/auth — lenient, CI is authoritative).
+    """
     try:
         out = subprocess.run(
             ["gh", "issue", "view", num, "--json", "state", "--jq", ".state"],
@@ -43,10 +49,14 @@ def gh_issue_state(num: str) -> str | None:
             timeout=30,
         )
     except (OSError, subprocess.SubprocessError):
-        return None
-    if out.returncode != 0:
-        return None
-    return out.stdout.strip().upper() or None
+        return "TRANSIENT"
+    if out.returncode == 0:
+        return out.stdout.strip().upper() or "TRANSIENT"
+    # gh ran but failed — is it "no such issue" (typo, our fault) or infra?
+    err = (out.stderr or "").lower()
+    if "could not resolve" in err or "not found" in err or "no issue" in err:
+        return "NOT_FOUND"
+    return "TRANSIENT"
 
 
 def main() -> int:
@@ -65,9 +75,14 @@ def main() -> int:
     violations: list[str] = []
     for num in pending:
         state = gh_issue_state(num)
-        if state is None:
-            print(
-                f"  #{num}: state UNKNOWN (gh could not resolve) — not failing on this"
+        if state == "TRANSIENT":
+            print(f"  #{num}: gh transient failure — lenient (CI authoritative)")
+            continue
+        if state == "NOT_FOUND":
+            violations.append(
+                f"#{num} does not exist — a pending register row points at a "
+                f"non-existent issue (typo?). A row that can never resolve is "
+                f"silent-rot; fix the number or remove the row"
             )
             continue
         if state == "CLOSED":
--- a/scripts/check_pending_upgrades.py
+++ b/scripts/check_pending_upgrades.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+"""Pending-upgrade guard.
+
+Reads docs/foundations/pending-upgrades.md, finds every register row tagged
+`status: pending` with an `issue: #N`, and queries that issue's state via `gh`.
+FAILS (exit 1) if any pending row's unblocking issue is CLOSED — the owed
+upgrade must be applied and the row marked `resolved` before the build is green.
+
+This converts "met-pending" rulings from memory-dependent promises into a
+mechanical merge-gate: a closed unblocking issue with an un-resolved row cannot
+ship (PATTERN-01 — only an un-mergeable state holds).
+
+Skips gracefully (exit 0, loud note) if `gh` is unavailable / unauthenticated,
+so it never blocks a local commit on missing credentials — CI is the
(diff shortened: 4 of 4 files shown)


JamesPagetButler:
## Red Team Review — PR #526 (pending-upgrade guard)

**Tier 2 infra · Sabine (empiricism) / Grothendieck (structure) / Knuth (correctness)**

All probes empirically reproduced against head `c9bdec3`. Verdict tables first.

### Probe results (traffic-light)

|---|---|---|---|
| 1 | gh-unavailable skip graceful (exit 0, not false pass)? Does CI have GH_TOKEN so it won't skip? | `shutil.which("gh") is None → exit 0` with a loud `SKIPPED` note. CI step sets `env: GH_TOKEN: ${{ github.token }}`; the `gh` binary is preinstalled on `ubuntu-latest`, so `which("gh")` is non-None and the skip branch is NOT taken in CI. The skip is the right call for *local* runs only. | 🟢 |
| 2 | Regex `issue:\s*#(\d+)\s*\|\s*status:\s*(pending\|resolved)` parses rows, no prose false-match | Against the real register: matches exactly `[('525','pending')]`. Explanatory prose ("...its row is still status: pending") → **no match** (no `issue: #N \|` prefix). The HTML hint comment `issue: #N \| status: pending\|resolved` → **no match** (`#N` is not `#\d+`). | 🟢 |
| 3 | Workflow step has GH_TOKEN env | Present: `env: GH_TOKEN: ${{ github.token }}` on the step. | 🟢 |
| 4 | CLOSED issue + `resolved` row → passes (resolved rows ignored)? | Confirmed: `pending = [num for num,status in rows if status=="pending"]` filters resolved rows out before any gh call. Resolved + closed → empty pending list → exit 0. | 🟢 |
| 5 | Typo'd / nonexistent issue → `gh` errors → state None → "not failing" — too lenient? | **Real risk, see below.** A wrong issue number silently never guards: `gh issue view` non-zero → `None` → `print("state UNKNOWN ... not failing")` → continue. A mistyped row is a permanently dead guard with no signal that the typo happened. | 🟡 |
| 6 | Runs on the right trigger paths | `lean-foundations.yml` adds both `scripts/check_pending_upgrades.py` and `docs/foundations/pending-upgrades.md` to `on.pull_request.paths`. | 🟢 |

### Knuth (correctness) — probe 5, the one substantive flag

The `state is None → continue (not failing)` branch is defensible for **transient** gh failures (network blip, rate limit) — failing CI on a flaky API call would be worse. But it conflates two distinct conditions:

- **Transient failure** (network/auth) — lenient-continue is correct.
- **Permanent failure** (issue number is a typo / never existed) — lenient-continue means the row is a **dead guard forever**, silently. The whole point of the register is to make "we'll update later" un-forgettable; a typo'd issue number reintroduces exactly the silent-rot failure mode (PATTERN-01) the PR exists to kill, just one level up.

This is **not BLOCKING** for an infra guard with a single hand-curated row (#525 verified to exist and be OPEN — the live row works today). But it is a real latent gap. Recommended hardening (defer to housekeeping, non-blocking):
- Distinguish "issue not found" (gh exit + stderr `Could not resolve`/`not found`) from generic failure, and treat *not-found* as a violation (a typo should be loud), while keeping transient failures lenient. Or, more simply: a one-time existence check that a row's `#N` resolves at all, failing on unresolvable numbers.

### Grothendieck (structure)

Clean separation: register (data) / guard (mechanism) / workflow (trigger). The guard is pure-functional over the register text + gh state; resolved rows are a forensic record, not deleted — good provenance hygiene. The "only an un-mergeable state holds" framing is structurally sound: a closed issue + pending row is a hard RED, which is the correct lattice position (you cannot ship the promise un-kept).

### Sabine (empiricism)

The PR's own evidence claims ("exits 0 now since #525 OPEN; would exit 1 on a closed unblocking issue") reproduce. The regex, resolved-row-skip, and prose-non-match were all run, not asserted. Mechanically honest.

PR is process-tooling for the "AC met-pending-#N" mechanism itself (refs #507, #474/#522/#525); no linked issue carries a formal `- [ ] **AC:**` checklist for this guard. The PR's stated DoD (guard exits 0 now; exits 1 on closed unblocking issue; YAML valid; black/mypy clean) is internally verified and reproduces. **No unmet AC.**

### Verdict: **APPROVE**

Mechanically sound. One non-blocking edge-case (probe 5: typo'd issue → silent dead guard) worth a housekeeping issue, but the live register row is correct and the guard works as specified. The lenient-on-transient-failure posture is the right default for CI; only the permanent-typo sub-case is under-guarded.

JamesPagetButler:
## Gemini Review — PR #526 (pending-upgrade guard)

> **Provenance:** Generated via the **direct `gemini -m gemini-3-pro-preview` CLI** (MCP `review_document` unavailable this session). Posted verbatim — Claude did not author Gemini's review. Personas: Furey + Feynman.

---

### Tier-2 Infrastructure Review (Gemini Pair: Furey & Feynman)

| Category | Status | Component | Notes |
| :--- | :---: | :--- | :--- |
| **Logic & Regex** | 🟢 PASS | `check_pending_upgrades.py` | Regex is structurally sound; correctly filters out HTML hints and safely bypasses `resolved` rows. |
| **Environment** | 🟢 PASS | CI Workflow | `gh` is pre-installed on `ubuntu-latest`, and the step correctly injects `GH_TOKEN`. The local `which(gh)` bypass is safe for local dev. |
| **Permissions** | 🟡 CONCERN | `.github/workflows/lean-foundations.yml` | Strict `contents: read` drops issue read access. `gh` API calls may fail depending on repo visibility settings. |
| **Triggers** | 🟡 CONCERN | `.github/workflows/lean-foundations.yml` | Trigger path restriction defeats the operational purpose of a global merge gate. |
| **Error Handling** | 🔴 BLOCKING | `check_pending_upgrades.py` | Silent failure on unrecognized issue numbers introduces a structural vulnerability. |

**Verdict: 🔴 BLOCKING**

***

### Furey: Structural & Algebraic Rigor

The core mechanism of mapping a registry state to a physical constraint is conceptually sound, but the implementation contains a topological hole in its error handling. 

*   **The Regex (Probes 2 & 4):** Solid. `\d+` protects us from matching the `issue: #N` placeholder in the HTML comment. The list comprehension `[num for num,status in rows if status.lower()=="pending"]` ensures that a CLOSED issue with a `resolved` row evaluates to a null-op. The state machine behaves correctly under nominal conditions.
*   **The Null-State Vulnerability (Probe 5) [BLOCKING]:** The `state is None` branch silently executes a `continue`. If a developer typos an issue number (e.g., `#9999999`), `gh` returns a non-zero exit code, `gh_issue_state` returns `None`, and the guard effectively ignores the row. The registry rots silently.
    *   *Correction required:* An unresolvable issue reference in the registry is an invalid registry state. If `state is None`, it must be treated as a violation: `violations.append(f"#{num}: UNKNOWN or unreachable (typo?)")`.

### Feynman: Operational & Physical Reality

Let's look at how this machinery operates on the factory floor. The goal is to stop people from merging code if they've left an IOU hanging. 

*   **The Local Escape Hatch (Probe 1 & 3):** Exiting `0` when `gh` isn't installed is the right pragmatic call for a developer's local machine. We know the CI runner (`ubuntu-latest`) has `gh` baked into the image, and the `GH_TOKEN` is explicitly piped in, so the CI constraint holds firm.
*   **The Path Restriction Blindspot (Probe 6) [CONCERN]:** You've restricted the workflow to trigger only on changes to `scripts/check_pending_upgrades.py` and `docs/foundations/pending-upgrades.md`. Think about what happens in reality: Alice closes Issue #525 on Tuesday. On Wednesday, Bob opens an unrelated PR modifying `src/foo.py`. Because Bob didn't touch the registry markdown, the workflow doesn't trigger. The merge gate never drops! To be a true guard, it must run on *all* PRs so that *any* open PR acts as a tripwire for resolved IOUs.
*   **The Token Scope (Probe 7) [CONCERN]:** You explicitly defined `permissions: contents: read`. In GitHub Actions, explicitly declaring a permission sets all undeclared scopes to `none`. While public repositories sometimes permit anonymous reading of issues, passing a token with `issues: none` to the `gh` CLI can trigger API rejections. You need to explicitly add `issues: read` to the workflow permissions block to guarantee physical access to the GitHub API. 

**Summary for the Author:**
1. Change the `state is None` logic to trigger a violation, not a `continue`.
2. Remove the path restrictions from the workflow so it runs on all PRs.
3. Add `issues: read` to the workflow permissions.

JamesPagetButler:
## Conflict resolved by the PR's own thesis — Gemini's blocker adopted (fix pushed)

Reviewer split (Red Team non-blocking, Gemini BLOCKING on probe-5) resolves without a tie-break: a guard whose purpose is killing silent-rot may not contain a silent-rot hole. All three hardenings applied:

- **probe-5 (the blocker):** NOT_FOUND (typo'd/nonexistent issue) is now a loud violation, distinguished from TRANSIENT gh failure (network/auth, still lenient). Self-tested: bogus number → NOT_FOUND → violation; #525 → OPEN → held.
- **probe-6:** guard moved to its own all-PR workflow `pending-upgrades.yml` — runs on EVERY pull request, so it trips the moment an unblocking issue closes regardless of what the PR touches. Delivers the 'un-skippable' guarantee the body claimed.
- **probe-7:** explicit `issues: read` permission.

Delta re-review follows the fix commit.

JamesPagetButler:
## Red Team Review — Round 2 (delta) — PR #526

**Provenance:** Red Team personas (Sabine / Grothendieck / Knuth), Claude (claude-opus-4-8). Round-2 *delta* review of fix commit `bc52606` against round-1 reviewed base `c9bdec3`. Scope: the three changes addressing Gemini's round-1 BLOCKING (probe-5) plus the two probe-6/probe-7 hardenings. Round-1 Red Team verdict was APPROVE (non-blocking flags); this delta does not reopen settled ground.

### Delta under review

| Change | File | Probe |
|---|---|---|
| `gh_issue_state` → 4-valued (OPEN/CLOSED/NOT_FOUND/TRANSIENT); main() fails on NOT_FOUND & CLOSED-pending | `scripts/check_pending_upgrades.py` | probe-5 (the blocker) |
| Guard moved to own `pull_request:`-all-paths workflow | `.github/workflows/pending-upgrades.yml` (new) | probe-6 |
| Old foundations-scoped step removed | `.github/workflows/lean-foundations.yml` | probe-6 |
| `permissions: issues: read` on new workflow | `.github/workflows/pending-upgrades.yml` | probe-7 |

### Traffic-light verification

| # | Check | Status | Evidence |
|---|---|---|---|
| 1 | NOT_FOUND → genuine violation | 🟢 PASS | `main()` appends to `violations` on `state == "NOT_FOUND"`; non-empty `violations` ⇒ `return 1`. Confirmed loud + build-failing. |
| 1a | not-found vs transient distinction sound | 🟢 PASS | Live `gh issue view 99999999` returns rc=1, stderr `GraphQL: Could not resolve to an issue or pull request...`. Lowercased contains `could not resolve` → NOT_FOUND. Substring set `{could not resolve, not found, no issue}` covers the real error. |
| 1b | no false-TRANSIENT that resurrects the hole | 🟢 PASS | A genuine not-found ALWAYS emits "Could not resolve to an issue" on the GraphQL path → matched. Auth failure ("HTTP 401"/"gh auth login") does NOT contain those substrings → TRANSIENT → lenient. Acceptable: auth failure is infra, and CI uses `github.token` (always authed). A genuine not-found cannot be misclassified as transient. |
| 2 | new workflow triggers on ALL PRs | 🟢 PASS | `on: pull_request:` with no `paths:` filter. |
| 2a | old foundations step removed | 🟢 PASS | `grep -i pending` on `lean-foundations.yml` → zero hits. Both the trigger-path entries and the run-step gone. |
| 3 | issues:read present | 🟢 PASS | Explicit `permissions: contents: read / issues: read` block. |
| 4 | no regression | 🟢 PASS | `ROW_RE` unchanged; register parses to `[('525','pending')]`, resolved rows excluded by the `status=='pending'` filter; `shutil.which("gh") is None` graceful-skip path intact; both YAMLs `yaml.safe_load` clean; `py_compile` clean. |
| 5 | no NEW risk from the all-PR move | 🟢 PASS | Per-PR cost is one `setup-python` + one `gh issue view` per pending row (currently 1) — trivial. Uses `github.token`, no secret. No new attack surface (`issues: read` is least-privilege). |

### Adversarial probes (round-2-specific)

- **Knuth (silent-rot symmetry):** The fix correctly closes the dual of the original hole. Round-1 hole = closed-issue-not-noticed; the fix's own new hole would have been typo'd-issue-never-guards. Both now fail loud. The TRANSIENT bucket is the *only* remaining lenient path and it is correctly scoped to genuine infra failure. No residual silent-success path.
- **Sabine (does it actually gate?):** With the all-paths workflow, a closed unblocking issue trips the guard on *any* PR — the "un-skippable the moment #NNN closes" guarantee now actually holds, which the foundations-scoped trigger did not deliver. This is a real correctness improvement, not cosmetic.
- **Grothendieck (classification completeness):** The 4-valued return is now total over gh outcomes: rc==0+output → state; rc==0+empty → TRANSIENT (defensible — empty is anomalous, lenient is safe); rc!=0+resolve-error → NOT_FOUND; rc!=0+other → TRANSIENT; exception → TRANSIENT. No gap.

### Minor (non-blocking) observations

- **N1 (nit):** TRANSIENT on rc==0-but-empty-stdout is theoretically lenient, but `gh ... --jq .state` returning rc=0 with empty output is not a reachable real state for an existing issue; acceptable.
- **N2 (cosmetic):** The two-line YAML comment on `issues: read` references "the repo ever going private" — `issues: read` is good hygiene regardless; comment is accurate, no action.

These are not blocking and need not be fixed in this PR.

Round-1 verified the AC set (register + guard + graceful-skip) as PASS. This delta strengthens the guard without altering the AC surface; the round-1 BLOCKING raised by Gemini (probe-5: typo'd/nonexistent issue → silent dead guard) is now closed — NOT_FOUND is a loud, build-failing violation.

### Verdict: **APPROVE**

The fix resolves the round-1 blocker correctly and completely, the two hardenings (all-PR trigger, explicit issues:read) are sound, and no regression or new risk is introduced. Gemini RAISED probe-5 and must confirm its own sign-off.

JamesPagetButler:
## Gemini Review — Round 2 (delta) — PR #526

**Provenance:** Gemini (`gemini-3-pro-preview`) via CLI (MCP unavailable this session); personas Furey / Feynman. SEQUENTIAL after Red Team. Reviewing fix commit `bc52606` against round-1 base `c9bdec3`. Gemini raised the round-1 BLOCKING (probe-5), so this sign-off is load-bearing. Output below is verbatim.

---

Here is my review of PR #526 as the Gemini reviewer (Furey/Feynman personas):

**1. Probe-5 Blocker Resolution (NOT_FOUND vs TRANSIENT)**
Your fix properly separates a network/auth failure (where we want to fail open, since CI is authoritative) from a legitimate `gh` response that the issue doesn't exist (where we MUST fail closed to prevent silent rot). 
- **Is NOT_FOUND genuinely a build-failing violation?** Yes, it is now appended to the `violations` list and triggers an `exit 1` at the end of the script.
- **Can a genuine NOT_FOUND be misclassified as TRANSIENT?** It's highly unlikely. You've checked against three common `gh` error substrings (`"could not resolve"`, `"not found"`, `"no issue"`). If the `gh` CLI changes its GraphQL error message dramatically in the future, it would degrade to TRANSIENT (lenient), but for current and foreseeable `gh` versions, this is robust. 
- **Is lenient-on-auth-failure acceptable?** Yes. Local developers shouldn't be blocked from committing just because they aren't logged into `gh`. The authoritative check happens in GitHub Actions, which is injected with a valid `GH_TOKEN`.

**2. Probe-6 (The "Un-skippable" Guarantee)**
Yes, moving this to its own `.github/workflows/pending-upgrades.yml` with an unfiltered `on: pull_request:` trigger is exactly the right architectural move. If it were scoped to foundations-only paths, a random documentation PR could slip through while a pending issue was closed, violating the guarantee. The diff confirms the step is fully ripped out of `lean-foundations.yml`.

**3. Probe-7 (`issues: read` Permission)**
Spot on. Explicitly stating `issues: read` ensures the `gh issue view` command will have the token scope it needs regardless of the repository's default permission settings or future visibility changes. 

**4. Regressions**
I see no regressions:
- `ROW_RE` parsing is untouched and solid.
- `resolved` rows are still correctly ignored (they don't make it into the `pending` list).
- The `shutil.which("gh") is None` check remains intact for graceful skipping on systems without the CLI.
- The YAML is structurally valid.

**5. New Risks**
No significant new risks. Running a fast, lightweight Python script on every PR consumes negligible CI minutes. The permissions are properly scoped to read-only access. 

My Round-1 probe-5 blocker regarding silent rot from missing/typo'd issue numbers is **RESOLVED**.

**VERDICT: APPROVE**

---

PR #529, opened 2026-06-08
[infra] close union-merge edge-5: aggregator stale-import guard (#527 Gemini point-5)

JamesPagetButler:
## Summary

Post-merge hardening of the union-merge driver (#527), closing the one residual its Gemini review flagged **🟡 MONITOR** (point 5).

**The gap:** union-merge has two failure modes. The duplicate-import guard (already in #527) catches mode 1 (identical doubled line). Mode 2 — a **divergent same-line edit** (two branches rename the same import differently) → union keeps *both* lines, one possibly pointing at a deleted/renamed module — produces **two different** lines, which the dup-guard misses. That left `lake build`'s missing-module error as the *only* net: slow, and the fast linter had no check that aggregator imports resolve to real files.

**The fix:** `check_layer_imports.py` now verifies **every `import QBP.Foundations.X` resolves to an existing `Foundations/X.lean`** — the inverse of the existing completeness check (files→imports), giving imports→files. Fast, deterministic, companion to the dup-guard; together they cover both union failure modes. Bonus: also catches plain stale/forgotten imports (delete a file, forget its aggregator line).

Self-tested: clean tree passes; an injected `GhostModule` import is flagged loudly with the union-divergent-edit hint.

## Theory-element headline
N/A — CI tooling; no physical claim.

## CTH anchor impact
- **New anchors:** none.
- **Routing axis**:
  - [x] N/A — merge/lint tooling, no anchor-bearing paths

- [x] clean tree: `layer imports clean`
- [x] stale-import injection caught (self-test in commit msg)
- [x] CI green — all checks pass + mergeable; both APPROVE @ 2026-06-08

## Known-benign corner (reviewer-noted, non-defect)
If two branches rename the *same* import to two *different* existing modules (Foo→Bar, Foo→Baz; both files exist, Foo deleted), union keeps both `import Bar` + `import Baz` — both resolve, neither guard fires, the aggregator names two valid modules where one was intended. Ruled benign by both reviewers: build passes, the extra import is at worst a too-wide surface, and detecting it needs rename-intent reconstruction not recoverable from the file tree. **This guard + the dup-guard fully cover the two BROKEN-BUILD modes Gemini #527 point-5 named** (identical-dup; stale/missing rename); this corner is build-safe.

Refs #527 (Gemini point-5), #507 (gate family).


JamesPagetButler:
## 🔴 Red Team Review — PR #529 (stale-import guard)

**Tier 2 infra · Lead: Knuth · Panel: Sabine, Grothendieck**
**Reviewed against HEAD `a55f929`. All probes run locally.**

---

### Verdict table

| Reviewer | Domain | Verdict |
|---|---|---|
| **Knuth** | correctness / mechanical | ✅ APPROVE |
| **Sabine** | does-it-actually-do-the-thing | ✅ APPROVE |
| **Grothendieck** | completeness of the close | ✅ APPROVE-WITH-NOTE |

**Panel verdict: APPROVE.** One honest residual noted (probe-3), ruled benign / out-of-scope.

---

Linked: #527 (Gemini review point-5, 🟡 MONITOR), gate family #507. No formal `- [ ] **AC:**` checklist on a dedicated issue; the binding criterion is the PR's own contract: *"every `import QBP.Foundations.X` resolves to an existing `Foundations/X.lean`, else flag."*

| Criterion | Assessment | Evidence |
|---|---|---|
| imports→files resolution check added | **PASS** | diff adds the `for imp in sorted(imported)` loop |
| catches mode-2 union artifact (divergent rename → stale line) | **PASS** | ghost-import injection → exit 1, correct message (probe-3 negative test) |
| no false positives on non-Foundations imports | **PASS** | `if not imp.startswith("QBP.Foundations."): continue` |
| clean on current aggregator | **PASS** | `layer imports clean`, exit 0, on both master tree and PR-branch tree |

---

### Probe results (Knuth — I ran every one)

**Probe 1 — path resolution correctness.** `PROOFS / Path(*imp.split(".")).with_suffix(".lean")` with `PROOFS = .../proofs`:
- `QBP.Foundations.Foo` → `proofs/QBP/Foundations/Foo.lean` ✅
- Nested dotted: `QBP.Foundations.Sub.Bar` → `proofs/QBP/Foundations/Sub/Bar.lean` ✅ — the `*imp.split(".")` splat handles arbitrary depth correctly.
- Repo reality: `find proofs/QBP/Foundations -type d` shows **no subdirectories** — all 16 modules are flat (`ArtinCore` … `Octonion32Count`). So no `X.Y` modules exist today, but the mapping is correct *if* one is ever added. No latent bug. ✅

**Probe 2 — no false positives.** Guard is gated on `imp.startswith("QBP.Foundations.")`. Mathlib / `Init` / cross-layer imports are skipped. Verified clean on the real aggregator (16 imports, all resolve). The `sorted()` over a `set` makes output order deterministic — good, error messages won't churn across runs. ✅

**Probe 4 — mechanical.** `black --check` → unchanged. `mypy` → *Success: no issues found*. CI: `.github/workflows/lean-foundations.yml` job `foundations-standard` already runs `python3 scripts/check_layer_imports.py` (line 43), and `scripts/check_layer_imports.py` is in the path-filter trigger list (line 25). **No workflow change needed — confirmed.** ✅

---

### Sabine — does it actually fire?

I don't trust a guard I haven't watched trip. Injected `import QBP.Foundations.GhostModule` into the aggregator on the PR-branch worktree:

```
LAYER-IMPORT VIOLATIONS:
  - QBP/Foundations.lean: import QBP.Foundations.GhostModule resolves to no file
    (QBP/Foundations/GhostModule.lean missing — stale/renamed import,
     possibly a union-merge divergent-edit artifact)  → exit 1
```

Fires, exits non-zero, message names the file and the union-divergent hypothesis. Real net, not a paper net. ✅

---

### Grothendieck — completeness of the close (probe-3 ruling)

The claim is that this guard + the existing dup-guard cover **both** union-merge failure modes Gemini named:
- **Mode 1** (identical doubled `import Foo`) → caught by dup-guard (in #527). ✅
- **Mode 2** (branch A renames `Foo→Bar`, branch B renames `Foo→Baz`; union keeps two *different* lines; dup-guard blind) → if either rename target is missing, **this** guard fires on the stale line. ✅

**The residual the prompt asks me to rule on:** what if branch A does `Foo→Bar` and branch B does `Foo→Baz`, *both* `Bar.lean` and `Baz.lean` exist, and `Foo.lean` is deleted by both? Union → aggregator imports **both** `Bar` and `Baz`; both resolve; neither guard fires. The aggregator now names two modules where the human intent was one.

**Ruling: this is a benign residual, and it does not block.** Reasoning:
1. Both `Bar` and `Baz` are *valid, existing* Foundations modules. The aggregator's contract is "every Foundations module is reachable / no dangling import" — both halves still hold. `lake build` succeeds.
2. The "extra import" is semantically harmless in Lean: importing a real module that compiles adds nothing incorrect to the build; it's at worst a slightly-too-wide import surface, identical in kind to a human adding one redundant `import` line by hand.
3. Detecting it would require *intent* reconstruction (knowing `Bar` and `Baz` are "the same module under two names"), which is not recoverable from the file tree — there is no artifact that says they were once one file. No static linter can catch it without a rename-provenance oracle. It is out of scope for a file-resolution check, by construction.
4. It is strictly less dangerous than the modes that *are* caught: those produced broken builds; this produces a correct build with one redundant valid import.

So: a residual exists, it is real, it is benign (correct build + valid modules), and it is fundamentally outside what a resolves-to-a-file check can address. The close is complete **with respect to the two failure modes Gemini named** (broken-build artifacts), which is exactly the scope of point-5. I'd note it in the PR body as a known-benign corner for honesty, but it is not a defect and must not gate.

---

### Summary

Tight, well-commented, single-purpose hardening. Path math correct (incl. the nested case that doesn't exist yet but would work). No false-positive surface. Self-test verified live. CI wiring already present. The one theoretical residual is benign-by-construction and out of a file-resolver's reach. **APPROVE.**

JamesPagetButler:
## 🔵 Gemini Review — PR #529 (stale-import guard)

**Tier 2 infra · Sequential (after Red Team / Knuth) · Reviewers: Furey + Feynman**

> **Provenance:** Real Gemini API call. Model `gemini-3-pro-preview` via `gemini` CLI 0.45.2 (`--approval-mode plan`, read-only). Input = review brief with the diff, the probe-3 residual question, and Red Team's verified-facts list. Output reproduced verbatim below; Claude did not author any part of Gemini's half.

---

**VERDICT: APPROVE**

**FUREY (Structural Analysis):**
The logic introduced correctly maps the import namespace (`QBP.Foundations.*`) to the filesystem path (`PROOFS / Path(*...).with_suffix(".lean")`) and validates file existence. This closes the structural gap where a union-merge could silently introduce an import for a non-existent file due to divergent renames or deletions.

Regarding the probe-3 residual case (Branch A: Foo->Bar; Branch B: Foo->Baz; both files exist): I concur with the Red Team. In this scenario, the aggregator correctly reflects the union of both valid modules. A file-resolver's invariant is simply "do all requested imports exist?" It cannot deduce the semantic intent that only one of Bar or Baz should exist without a higher-level state tracking mechanism (such as an overarching rename provenance oracle). If both modules are structurally sound and exist on disk, the union-merge operated correctly on the set of modules. Any logical redundancy or conflict introduced by both modules coexisting is a matter for the Lean compiler, not a defect in this CI linter.

**FEYNMAN (Pragmatic Verification):**
Does it actually work? Yes. The Red Team verified the guard catches injected ghosts and leaves the good files alone. The path mapping is solid, and the `startswith` gate ensures it ignores external libraries like Mathlib so we don't get false alarms. No workflow changes are needed, and it passed all the black and mypy checks.

As for the residual case: it's perfectly fine. If both files are there, Lean compiles, and the linter says "everything you asked for exists", then the machine is doing exactly what it's supposed to do. If we have two modules where we only wanted one, the Lean compiler will likely complain about duplicate theorems, or we just end up with some extra code. But this script is a map-checker, not a mind-reader. Trying to fix that here would mean building an unnecessarily complex state machine. We shouldn't build a Rube Goldberg machine for a theoretical edge case. Ship it.

PR #532, opened 2026-06-08
[infra] authenticate lake cache fetch — fix 'failed to parse latest release tag'

JamesPagetButler:
## Summary

The `lake build` job's Mathlib cache fetch hit GitHub's API **unauthenticated** (60 req/hr, shared per runner IP). Today's heavy CI load (dozens of reruns across the foundation wave) drained that quota; the cache tooling then gets a rate-limit body instead of a release tag and dies with the opaque **`failed to parse latest release tag`**. The tell: earlier PRs (#528 at 9m7s) built green, later ones (#531) failed — rate-limit exhaustion, not a code/cache-version fault.

**Fix:** pass `github.token` (as both `GH_TOKEN` and `GITHUB_TOKEN`) to the cache-get and build steps → 5000 req/hr authenticated → doesn't exhaust. Affects **all** foundation lake-builds, including the milestone #531 (octonion_artin / AC2).

This PR's own `lake build` job is the test: if it goes green (authenticated cache fetch succeeds), the fix works.

## Theory-element headline
N/A — CI infra; no physical claim.

## CTH anchor impact
- **New anchors:** none.
- **Routing axis**:
  - [x] N/A — CI workflow only

## Mechanical verification evidence
- [x] YAML valid
- [x] **This PR's lake-build job green = the fix verified** (authenticated fetch)
- [x] CI green overall

Refs #531 (milestone blocked by this), the recurring-flake hardening family (#530).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #531, opened 2026-06-08
[foundations] #474 PR-C2d: octonion_artin — FULL Artin at 𝕆 (closes AC2)

JamesPagetButler:
## Summary

The final Artin chunk. **`octonion_artin`** proves the full theorem at 𝕆: every subalgebra of CDAlg ℝ 3 generated by two elements is associative — `∀ u v w ∈ NonUnitalAlgebra.adjoin ℝ {x,y}, (u·v)·w = u·(v·w)`. Stated over Lean's genuine generated-subalgebra object (not a monomial list); the strongest faithful form. **This closes #474 AC2.**

Beekeeper-ruled scope (FULL Artin, not the alternativity-only scoping) — delivered, kernel-checked, **from alternativity via the span route, NO Hurwitz/classification appeal** (W1 truth-in-labelling stated in-file; the 4-dim span is a *consequence* of C2c's 16-cell closure, never a premise).

### Assembly (all conceptual work pre-merged — 3 lines)
1. `mem_span4_of_mem_adjoin` (C2c) — adjoin-membership → span{1,x,y,xy}-membership for u,v,w
2. `assoc_vanishes_on_span4` (C2a) — assoc = 0 on that span
3. `rw [assoc, sub_eq_zero]` — vanishing gives re-association

No glue lemma needed (C2a/C2c share the `gen4` definition). Companions: `octonion_artin_assoc` (assoc=0 form, for the index `#check`) + `octonion_artin_gen` (concrete `(x*y)*x = x*(y*x)` witness, no hypotheses).

## Theory-element headline
N/A — pure-math foundation; no physical claim.

## CTH anchor impact
- **New anchors:** none.
- **Anchor-rule terminations:** [x] Lean file:line — octonion_artin + companions in Artin.lean
- **Routing axis**:
  - [x] theory-axis → co-sign by @qbp-oppenheimer needed (default for Tier 3)

## Mechanical verification evidence
- [x] `lake build QBP.Foundations` green (3132 jobs)
- [x] gates PASS; aggregator wired (union-merge live)
- [x] `#print axioms` all 3: {propext, Classical.choice, Quot.sound}
- [x] zero sorry / native_decide / decide / vacuous-True
- [x] CI green ← notifier verdict canonical

**Closes #474 AC2 (Artin).** With this, AC2's full obligation (Moufang ×3 + Hurwitz + Artin at 𝕆) is proven. Part of #474 Wave 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #533, opened 2026-06-08
[infra] link-check: PRs scan only changed .md; weekly cron full-sweep (closes #530)

### LINKED ISSUE #530: [infra] markdown-link-check: harden against transient connection flakes (Status 0 / ESOCKETTIMEDOUT)
Recurring CI flake class (2× now): #515 (webaim.org), #529 (github.com/KaijuEngine, even github.com/JamesPagetButler/QBP itself). The link-checker reports **Status 0 / ESOCKETTIMEDOUT** — a *connection-level* failure, never a real dead-link (404) signal — and fails the PR, forcing manual reruns. A Status 0 on github.com is runner network noise, not a broken link.

Per the session's recurring-failure → structural-fix discipline, the rerun-each-time approach should be replaced by a structural fix. Options:
1. **CI step-level retry wrapper** (recommended): wrap the link-check step to retry 2-3× on failure (transient flakes clear; genuine dead links persist across retries).
2. Bump markdown-link-check  (default 10s) in mlc_config.json — reduces timeout flakes but doesn't eliminate connection failures.
3. Treat Status 0 as non-fatal (markdown-link-check has no clean config for this; would need a post-process filter).

**AC:**
- [ ] transient connection failures (Status 0 / ESOCKETTIMEDOUT) no longer fail a PR on the first flake
- [ ] genuine dead links (persistent 404) still caught
- [ ] no per-incident manual rerun needed

JamesPagetButler:
## Summary

Closes #530 — the recurring link-check flake (now **3×**: #515 webaim, #529/#532 github.com) that's currently **blocking the #531 Artin milestone**.

**Root cause:** the workflow scans *all* repo `.md` on every PR, so any PR is hostage to a transient external-link timeout (Status 0 / ESOCKETTIMEDOUT) in a file it never touched. #532 changes *zero* `.md` files yet fails on github.com timeouts in unrelated docs.

**Fix:** `check-modified-files-only` = `yes` on `pull_request` events → a PR only checks links in the `.md` it changed. A **weekly cron full-repo sweep** (Mondays 06:00 UTC) + the push-to-master full scan still surface link-rot in untouched files — just on a cadence, not as a flaky per-PR tax. (modified-only is a no-op on push/schedule, so those stay full-scan.)

Effect: YAML/Lean-only PRs (#532, #531, and most foundation PRs) get a trivially-passing link-check; genuine dead links in changed docs are still caught immediately; repo-wide rot is caught weekly.

## Theory-element headline
N/A — CI infra; no physical claim.

## CTH anchor impact
- **New anchors:** none.
- **Routing axis**:
  - [x] N/A — CI workflow only

- [x] YAML valid
- [x] this PR changes 0 `.md` → its own modified-only link-check passes trivially (self-proving)
- [x] CI green — all checks pass + mergeable; both APPROVE; modified-only link-check self-validated (0 .md changed) @ 2026-06-08

Closes #530. Unblocks #531 (Artin/AC2) + #532 once they refresh.

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -1,5 +1,15 @@
 # .github/workflows/link-checker.yml
-# This workflow checks all markdown files for broken links.
+# Link checking, scoped to avoid flaky-external-link noise blocking unrelated PRs.
+#
+# Recurring incident (3x: #515 webaim, #529/#532 github.com): a PR scanning ALL
+# repo .md files fails on a transient timeout (Status 0 / ESOCKETTIMEDOUT) of an
+# external link in a file the PR never touched. A YAML-only or Lean-only PR has
+# no business being blocked by link-rot elsewhere.
+#
+# Fix (issue #530): on PRs, check ONLY the .md files the PR changed
+# (check-modified-files-only). A weekly scheduled full-repo scan catches rot in
+# untouched files, so nothing is lost — link-rot is surfaced on a cadence, not
+# imposed as a flaky tax on every unrelated PR.
 name: "Link Checker"
 
 on:
@@ -9,6 +19,9 @@ on:
   pull_request:
     branches:
       - master
+  schedule:
+    # weekly full-repo sweep (Mondays 06:00 UTC) — catches rot in untouched files
+    - cron: '0 6 * * 1'
 
 jobs:
   markdown-link-check:
@@ -23,3 +36,7 @@ jobs:
           use-quiet-mode: 'yes'
           use-verbose-mode: 'yes'
           config-file: 'mlc_config.json'
+          # PRs: only the files they changed (kills unrelated-flake blocking).
+          # push-to-master / scheduled: full repo (modified-only is a no-op there,
+          # so the cron sweep still scans everything).
+          check-modified-files-only: ${{ github.event_name == 'pull_request' && 'yes' || 'no' }}


JamesPagetButler:
## 🔴 Red Team Review — PR #533 (Knuth seat, Tier-2 infra)

*Scope: `.github/workflows/link-checker.yml` — scope PR link-checks to changed `.md`, add weekly cron + retain push-to-master full sweep. Closes #530.*

I read the action's source (`gaurav-nelson/github-action-markdown-link-check@v1`, `entrypoint.sh` + `action.yml` + README) rather than reasoning from the docs alone, because the load-bearing claim here is a behavioral one about how the action computes "modified files."

### Verdict table

| # | Probe | Verdict | Evidence |
|---|-------|---------|----------|
| 1 | Ternary expression + input semantics | ✅ PASS | Standard GHA `A && B \|\| C` idiom; yields string `'yes'` on PR, `'no'` else. Action treats `"yes"` literally (`if [ "$CHECK_MODIFIED_FILES" = "yes" ]`). |
| 2 | No coverage loss (cron + push still full-scan) | ✅ PASS | On non-PR events the input is `'no'` → action takes the `find` full-tree branch. Cron + push both full-scan. |
| 3 | **fetch-depth / modified-detection** | ✅ PASS (not blocking) | Action self-fetches base: `git fetch origin master --depth=1` then `git diff` vs that hash. Independent of the checkout step's depth. |
| 4 | Self-proving (0 `.md` changed) | ✅ PASS | `git diff --diff-filter=AM` yields empty file list → loop is a no-op → trivially passes. |

**Overall: APPROVE. No BLOCKING findings.** Two NON-BLOCKING notes below.

---

### Probe 3 — the one that mattered (resolved in favor of the PR)

The concern was real and correctly prioritized: the checkout step is

```yaml
- uses: actions/checkout@v4   # no fetch-depth → shallow (depth 1)
```

If the action diffed against history that a shallow clone lacked, modified-detection could silently find **zero** files and pass everything (a false-green that would defeat the entire fix). I traced `entrypoint.sh` to settle it:

```sh
if [ "$CHECK_MODIFIED_FILES" = "yes" ]; then
   git config --global --add safe.directory '*'
   git fetch origin "${BASE_BRANCH}" --depth=1 > /dev/null     # ← self-fetch
   MASTER_HASH=$(git rev-parse origin/"${BASE_BRANCH}")
   ...
   mapfile -t FILE_ARRAY < <( git diff --name-only --diff-filter=AM "$MASTER_HASH" -- "${FOLDER_ARRAY[@]}")
```

**The action does its own `git fetch origin master --depth=1` and diffs the working tree against `origin/master`.** It does NOT depend on the workflow's checkout depth. The official README example likewise uses a bare `actions/checkout@master` with no `fetch-depth`. So the missing `fetch-depth: 0` is **not** a gap — adding it would be harmless but unnecessary. Probe 3 is **not blocking**.

### Probe 1 — `base-branch` defaults to `master` (silent dependency, OK here)

The PR does not set `base-branch`. The action default is `master` (`action.yml`), which matches this repo's base. So the omitted input is correct **by coincidence of the default**. If the repo's default branch were ever renamed (e.g. `main`), this line would silently diff against a non-existent `origin/master` and the `git rev-parse` would fail the step. NON-BLOCKING, but worth a one-line comment near the input pinning the assumption, or explicitly setting `base-branch: master`.

### Probe 2 / the "no-op on push" claim — verified, with one honest caveat

PR body says "modified-only is a no-op on push/schedule." Mechanically that's right *because the expression resolves to `'no'` on those events* (event-gated), NOT because modified-only would self-neutralize if it ran. The distinction matters: if someone later changed the expression to unconditionally `'yes'`, a push event would run the diff branch and `git diff` vs `origin/master` on a push *to* master would (correctly) be empty → it WOULD then silently check nothing on push. The current event-gated form is correct; the safety rests on the `github.event_name == 'pull_request'` guard, so don't loosen that guard. NON-BLOCKING — current code is correct.

### Diff-semantics footnote (informational, not a defect)

The action diffs HEAD-working-tree against `origin/master` tip directly (two-dot `git diff <hash>`), **not** against the merge-base. On a stale PR branch this can flag a `.md` that master changed but the PR didn't — i.e. it may check *slightly more* than the PR's own delta. That fails safe (over-checks, never under-checks) and cannot reintroduce the unrelated-flake problem for files genuinely untouched on both sides. Noting only so no one is surprised if a rebase changes the checked set.

---

### Recommendation

**APPROVE.** The fix is correct, minimal, and the one credible failure mode (shallow-checkout false-green) is closed by the action's internal self-fetch. Optional polish (non-blocking): set `base-branch: master` explicitly to harden against a future default-branch rename. CI-green box remains to be ticked on the PR.

— Knuth (Red Team)

JamesPagetButler:
## 🔵 Gemini Review — PR #533 (Furey/Feynman seat, Tier-2 infra)

> **Provenance:** Generated by the real Gemini API via CLI `gemini -m gemini-3-pro-preview` (gemini CLI v0.45.2), 2026-06-08. Sequential after the Red Team (Knuth) review. Verbatim output below — Claude did not author or edit Gemini's verdict. Reviewer was given the diff + verified `entrypoint.sh`/`action.yml`/README facts for `gaurav-nelson/github-action-markdown-link-check@v1` and the four probes.

---

### Verdict Table

|---|-------|---------|--------|
| 1 | Syntax & Intake | GHA ternary expression `${{ ... && 'yes' \|\| 'no' }}` is valid and yields the literal string `'yes'` on PRs. The action's bash script evaluates it directly with `if [ "$CHECK_MODIFIED_FILES" = "yes" ]`. The interface mapping is exact. | 🟢 PASS |
| 2 | Coverage | `push` and `schedule` events resolve to `'no'`. The action's logic reverts to its default full-tree `find` scan. This guarantees complete coverage and ensures no documentation escapes validation. | 🟢 PASS |
| 3 | Shallow Checkout | `actions/checkout@v4` on a PR checks out the *merge commit* as the working tree. Diffing this working tree against the fetched `origin/master` tip requires only the two tree objects, not historical traversal. A depth-1 fetch is structurally sufficient for a two-dot diff. The `AM` filter accurately extracts what was added or modified. No false greens. | 🟢 PASS |
| 4 | Zero Files Edge Case | With 0 `.md` files changed in the PR, the `git diff` output is empty. The action's file list is empty, its internal execution loop is skipped, and it exits with status 0. It is cleanly self-proving. | 🟢 PASS |

### Tier-2 Infra Review

The motivation for this change is structurally necessary: transient external network failures on unrelated documentation must not block the development velocity of the core programme, especially during critical milestones. 

The primary adversarial vector is whether the interaction between `actions/checkout@v4`'s shallow `depth: 1` clone and the action's internal self-fetch of `origin/master` creates a blind spot. It does not. Git does not need historical commit ancestry to compute a diff between two available trees. Because `actions/checkout` constructs the PR merge commit locally, its working tree contains all intended PR changes. When the action runs `git diff origin/master`, it compares the target `master` tree directly against the PR's working tree. Any file the PR added or modified is correctly identified by the `AM` filter. There is no mechanism here that would silently drop a modified `.md` file.

The edge case of zero `.md` files (as demonstrated by this YAML-only PR) is natively handled by the action skipping its iteration loop and returning a successful exit code, immediately unblocking the pipeline.

**One minor constraint to note:** The action explicitly fetches and diffs against `origin/master` by default. If the QBP federation ever begins running PRs against release or feature branches, this action will compute the diff against `master` rather than the true base branch, potentially resulting in massive over-scanning or missing files. Given that the current default branch is `master`, this is a theoretical constraint rather than an active defect, but it is worth documenting.

The weekly cron and `push` triggers correctly ensure the global state remains continuously validated. The logic is tight, the syntax is valid, and the risk of false positives is mitigated.

**APPROVE**

---

